PR #25, opened 2019-11-02
Several updates to interface

makaimann:
This pull request makes non-trivial changes to the API.
* It renames `make_term` for creating symbols to `make_symbol`
* It renames all `make_value` functions to `make_term`
* It adds a dedicated class with an internal cache for transferring terms to a new solver
* It removes the `has_symbol` and `lookup_symbol` functionality
* It make `make_sort` mirror the `make_term` functionality -- takes 1-3 sorts or a vector of sorts
* It uses explicit integer sizes instead of `unsigned int`, etc...


makaimann:
@lstuntz, once this is merged the API will be different. You should probably rebase your branch once these changes are in master. As long as you haven't started on the solver yet, there shouldn't be any conflicts.

include/solver.h
```diff
@@ -99,46 +99,46 @@ class AbsSmtSolver
...
-                         const std::vector<Sort> & sorts,
+                         const SortVec & sorts,
                          const Sort & sort) const = 0;
 
   /* Make a boolean value term
    * @param b boolean value
    * @return a value term with Sort BOOL and value b
    */
-  virtual Term make_value(const bool b) const = 0;
+  virtual Term make_term(const bool b) const = 0;
```
barrettcw: const bool -> bool

barrettcw: Look for other ints/bools with const qualifiers (e.g. make_sort) - these should be removed

makaimann: Thanks for the feedback/reminders! I made the changes in recent commits (looks like you already looked at those).

PR #28, opened 2020-01-29
Added Yices Sort functionality.

amaleewilson:
Trying to make an incremental pull request. These files contain everything for Yices Sorts.

### DIFF
--- a/yices2/src/yices2_sort.cpp
+++ b/yices2/src/yices2_sort.cpp
@@ -1,0 +1,141 @@
+#include <sstream>
+
+#include "yices2_sort.h"
+
+#include "exceptions.h"
+
+using namespace std;
+
+namespace smt {
+
+// Yices2Sort implementation
+
+std::size_t Yices2Sort::hash() const
+{
+  // type_t is a unique id, see Yices2 docs.
+  return type;
+}
+
+uint64_t Yices2Sort::get_width() const
+{
+  size_t out_width;
+  if (yices_type_is_bitvector(type)) 
+  {
+    return (unsigned int)yices_bvtype_size(type);
+  }
+  else
+  {
+    throw IncorrectUsageException("Can only get width from bit-vector sort");
+  }
+}
+
+Sort Yices2Sort::get_indexsort() const
+{
+  // Arrays are functions in Yices.
+  if (yices_type_is_function(type))
+  {
+    return Sort(new Yices2Sort(yices_type_child(type, 0)));
+  }
+  else
+  {
+    throw IncorrectUsageException("Can only get index sort from array sort");
+  }
+}
+
+Sort Yices2Sort::get_elemsort() const
+{
+  // Arrays are functions in Yices.
+  if (yices_type_is_function(type))
+  {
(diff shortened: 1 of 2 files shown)


yices2/src/yices2_sort.cpp
```diff
@@ -0,0 +1,141 @@
...
+  else if (yices_type_is_bitvector(type))
+  {
+    return BV;
+  }
+  else if (yices_type_is_function(type))
+  {
+
+    // Test if array or actually function.
+    // This may not be the most efective way to do this. 
+    if (!is_function)
```
makaimann: Just thinking ahead to future PRs: I see that `is_function` is provided to the constructor, but how do you know which it is? If you can query the type for it, I'd rather you just compute it in the constructor (as opposed to passing it).

yices2/src/yices2_sort.cpp
```diff
@@ -0,0 +1,141 @@
...
+  }
+  else if (yices_type_is_bitvector(type))
+  {
+    return BV;
+  }
+  else if (yices_type_is_function(type))
+  {
+
+    // Test if array or actually function.
+    // This may not be the most efective way to do this.
```
makaimann: minor typo -- and if you decide to change how this is done, you can remove this comment.

yices2/src/yices2_sort.cpp
```diff
@@ -0,0 +1,141 @@
...
+      return ARRAY;
+    }
+    else
+    {
+      return FUNCTION;
+    }
+  }
+  else
+  {
+    throw NotImplementedException("Unknown Yices2 type.");
```
makaimann: Can yices provide a string representation of the type? If so, it might be nice to include it in the exception.

yices2/src/yices2_sort.cpp
amaleewilson: is_function exists only to distinguish between arrays and functions. It’s not possible to query the Yices API for whether this sort is an array type, because arrays are represented as functions.

Re how to know which it is, the solver’s make_sort is called with e.g. an array or function type for the sort. If it is an array, is_function is set to false, and if it’s a function, it’s set to true. 

I think it might be possible to remove this entirely, since everything “under the hood” isn’t really relying on this distinction.

makaimann: Gotcha, that's interesting. So yices doesn't even distinguish between "updateable" functions and uninterpreted functions. I'm wondering how it works if you iterate over terms. Because, when you get the function/array from term traversal, you don't know which it is (except for how it's used). Do you just mark it during term traversal?

amaleewilson: For types (not terms), Yices does not appear to distinguish between updatable and uninterpreted functions. For terms, there’s function or uninterpreted type which can be a function. 

For term traversal, are you referring to term iteration specifically or term transfer?

makaimann: Either really, when you get a function/array from term iteration, it should be marked correctly (in case you transfer it to another solver, or just want to know for some other purpose). From offline discussion, this sounds like the best way to do it!

PR #30, opened 2020-01-31
Add yices2 support: Term and TermIter functionality

amaleewilson:
Not sure why it's saying that Sort files were changed, but they're the same as they were earlier. 

yices2_term.h and yices2_term.cpp contain all the term wrapper stuff. I tried to include some descriptive comments that explain the strange traversals for bvsums, sums, products, and powers.


yices2/src/yices2_sort.cpp
```diff
@@ -0,0 +1,143 @@
...
+  else
+  {
+    throw IncorrectUsageException("Can only get element sort from array sort");
+  }
+}
+
+bool Yices2Sort::compare(const Sort s) const
+{
+  shared_ptr<Yices2Sort> ys = std::static_pointer_cast<Yices2Sort>(s);
+  if (type == ys->type)
```
makaimann: Suggestion: Here you could just use `return type == ys->type;` Sorry I missed that in the sort review.

makaimann:
For now, everything in this PR looks good. For the bigger picture though, I'm still a bit hung up on this `is_function` thing. I know I didn't have a very good example for you before, but I think I have one now. Consider this line from `msat-tests.cpp`: https://github.com/makaimann/smt-switch/blob/57dde11611562f761c9e6c2501cde7b50479a759/tests/msat/msat-tests.cpp#L65

I believe this is fine for the yices2 implementation because the yices types will match. However, what would happen if you changed the line to:
```
assert((*uf_app->begin())->get_sort()->get_sort_kind() == FUNCTION);
```

Because you're getting the function back as a child of the term, I don't think that term iteration would have identified that it needs to be a function, so `is_function` wouldn't be set. Am I wrong?

amaleewilson:
Oops, probably should have done a separate PR, but the most recent commit contains everything for the current Yices2 backend implementation. There are still several issues, the most glaring of which is probably the Yices config / context handling that was updated today.

makaimann:
No worries! I think it's fine to have it all here. I'll do a pass over it soon, but otherwise just leave it here. You can let me know when you think it's ready for another look.

makaimann:
Lets switch this PR to merge directly to master, I don't think we need a separate branch in-between anymore.

CMakeLists.txt
```diff
@@ -87,6 +86,20 @@ if (BUILD_MSAT)
 
 endif()
 
+# should we build yices2?
+option (BUILD_YICES2
+  "Should we build the libraries for boolector" OFF)
```
makaimann:
```suggestion
  "Should we build the libraries for yices2" OFF)
```

python/gen-smt-solver-declarations.py
```diff
@@ -57,13 +65,19 @@ def create_msat_interpolator():
     c_SmtSolver cpp_create_msat_interpolator "smt::MsatSolverFactory::create_interpolating_solver" () except +
 '''
 
+DECLARE_YICES2='''
+cdef extern from "yices2_factory.h":
+    SmtSolver cpp_create_yices2_solver "smt::Yices2SolverFactory::create" () except +
```
makaimann:
```suggestion
    c_SmtSolver cpp_create_yices2_solver "smt::Yices2SolverFactory::create" () except +
```

makaimann:
I think we also need to have a separate category for solvers that support term iteration, otherwise Yices2 fails some of the tests. I've attached a patch that should fix this: `git apply termiter-solver-enums.txt`


[termiter-solver-enums.txt](https://github.com/makaimann/smt-switch/files/4340720/termiter-solver-enums.txt)

makaimann:
And then to fix the python tests there's this one:

[yices-python-fix.txt](https://github.com/makaimann/smt-switch/files/4340803/yices-python-fix.txt)

CMakeLists.txt
amaleewilson: Oops!

yices2/include/yices2_solver.h
```diff
@@ -0,0 +1,94 @@
...
+  {
+    yices_free_config(config);
+    yices_free_context(ctx);
+
+    // TODO: Should probably find a good place to
+    // call yices_exit.
+    // yices_exit();
+  };
+  void set_opt(const std::string option, const std::string value) override;
+  void set_logic(const std::string logic) const override;
```
makaimann: Hey, sorry I had changed this in a recent PR. Could you only have one `set_logic` and not make it `const`? Then I think we should be good to go -- Travis almost completely passed this time haha.

makaimann: Thanks for the change @amaleewilso! I think it needs to be changed here also though, right? https://github.com/makaimann/smt-switch/blob/6c1b2f7eddce2be40f8d14fc4527ec4cdeb44acb/yices2/src/yices2_solver.cpp#L118

amaleewilson: Thanks, should be fixed.

yices2/src/yices2_factory.cpp
```diff
@@ -0,0 +1,25 @@
...
+  // Must initialize only once.
+  // Different instances of the solver will have
+  // different contexts.
+  if (!initialized)
+  {
+    yices_init();
+    initialized = true;
+  }
+
+  return std::make_unique<Yices2Solver>();
```
makaimann: Travis failed again, sorry :/ I changed SmtSolver to be a `shared_ptr`.
```suggestion
  return std::make_shared<Yices2Solver>();
```

PR #31, opened 2020-02-08
Python bindings

makaimann:
This PR would add Cython-based python bindings. This is integrated into the cmake build structure. The main files are the following:

* `enums.{pxd,pxi}`: are the enum declarations and python wrappers, respectively
* `smt_switch_imp.{pxd,pxi}`: are the declarations and python wrappers for:
  * `Op`
  * `Result`
  * `Sort`
  * `Term`
  * `SmtSolver`
* `gen-smt-solver-declarations.py`: a python script that automatically generates `smt_solvers.{pxd,pxi}` for the build based on which solvers are being built
* `smt_switch.pyx`: just a wrapper that includes the other files

Most of this infrastructure is pretty straightforward. However, there was one non-trivial problem which had to be handled and I'm open to other suggestions for addressing this.

Some solvers, such as boolector, do manual reference counting. This is implemented in the smt-switch btor backend in the RAII style using constructors and destructors. However, to decrement the reference count for a sort or term, boolector needs the solver (called `btor` at the boolector level). Thus, the solver has to be destructed _after_ all of the terms/sorts it created. In C++ this all works out fine. However, because Python is a garbage-collection language, it calls `__del__` on Python objects (which then calls the class destructors) when there are no references to it _at the Python level_. 

This can cause problems because sometimes the solver and its sorts/terms all go out of scope at the same time, and the solver is destroyed first. Although it introduces some overhead, @cdonovick and I decided the best approach was to include a reference to the solver in every Sort/Term at the Python level, therefore ensuring that the solver is destroyed last.

@cdonovick You can start experimenting with these bindings whenever you're ready. There are basic instructions in the README.md, but let me know if you have any trouble installing it. I'd recommend boolector and mathsat for now. There are some features still missing from CVC4 (had to make some changes to the new CVC4 API and I haven't gotten around to updating smt-switch accordingly yet). Just be warned that I haven't really tested it yet. I plan to write more tests, but there's already one test file for `pytest` in `tests/python` that you can look at to get started. And you can always look at the headers in the top-level `include` to see the interface. By the way, for indexed operators, you have to use `Op`, e.g. `Op(Extract, 4, 0)`.


makaimann:
@amaleewilson I have a few tests in `tests/python/test_bv.py` but I still need more. I thought I'd let you start reviewing whenever you have a chance though. Up to you if we wait for more tests before merging. I'd like to at least test ints, reals, ufs and arrays briefly. Probably don't need unit tests because we'll have those at the C++ level eventually.

makaimann:
Nevermind, I found some edge cases for Ops so I added unit tests for that.

README.md
```diff
@@ -41,3 +41,12 @@ where `./custom-cvc4/build/src/libcvc4.a` and `./custom-cvc4/build/src/parser/li
 
 # Building Tests
  You can build tests with `make test` from the build directory. After you have a full installation, you can build the tests yourself by updating the includes to include the `smt-switch` directory. For example: `#include "cvc4_factory.h"` -> `#include "smt-switch/cvc4_factory.h"`.
+
+## Debug
+The tests currently use C-style assertions which are compiled out in Release mode (the default). To build tests with assertions, please add the `--debug` flag when using `./configure.sh`.
+
+# Python bindings
```
amaleewilson: Maybe add how to setup cython or have a separate contrib/ script for it. I looked at the .travis.yml to figure out how to make it build.

amaleewilson: Also, it's probably not a README thing as much as a DEVELOPERS thing, but maybe include a quick note on running your pytests.

tests/python/test_unit.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import smt_switch as ss
+
+from collect_solvers import solvers, full_bv_support
+
+@pytest.mark.parametrize("create_solver", full_bv_support)
+def test_unit_op(create_solver):
```
amaleewilson: Python tests pass when I run them, but I think there needs to be more coverage.

python/smt_switch_imp.pxd
```diff
@@ -0,0 +1,125 @@
...
+
+from enums cimport SortKind
+from enums cimport PrimOp
+
+ctypedef shared_ptr[AbsSort] Sort
+ctypedef shared_ptr[AbsTerm] Term
+ctypedef shared_ptr[AbsSmtSolver] SmtSolver
+ctypedef vector[Sort] SortVec
+ctypedef vector[Term] TermVec
+# ctypedef unordered_map[Term, Term] UnorderedTermMap
```
amaleewilson: Not sure what this is/was for, but remove this comment if it isn't necessary.

amaleewilson (approved):
Looks great to me. I would suggest only a slightly clearer description of setup and more tests, which I think you're already working on.

README.md
makaimann: Addressed with f552b36e710a0a748983f359809d03ed5d49fe58. Does that look good to you? I just put it in the README: in case users want to make sure it installed correctly.

tests/python/test_unit.py
makaimann: Agreed. I'll add more tests soon.

python/smt_switch_imp.pxd
makaimann: Ope, good catch! Removed with c534d7afab3064227bc999ae41bd44e9cbf5370b. If you're curious, I was just trying to declare it like I did for all the others, but it didn't work because it requires a third template argument for the hash function. Instead, I just declared the C++ UnorderedTermMap class and it's emplace method (this is used for `substitute`)

tests/python/test_unit.py
makaimann: Added a few more tests in 4bdf66fb7f6f76012d0a2f7ef57f075e9954a864, 94bf63f33c1e4306d5ffcb409b0a9333e0544c69, and b0d1fb35131817168f1060a9bfce4b59693a355e

PR #37, opened 2020-02-12
Fix TermIter clone and assignment

makaimann:
I found a bug today in the generic `TermIter`. This came up because the Cython bindings will create a null `TermIter` and assign to it. The `clone` implementation was incorrect (it needs to rely on the specific backend solver implementations to be able to clone correctly). This never came up before, because all the tests never manipulated a `TermIter` directly (they were only used implicitly for iteration e.g. in `for (auto t : term)`).

@amaleewilson You'll need to implement the `clone` method for `Yices2TermIter`. It basically just returns a pointer to a copy. See https://github.com/makaimann/smt-switch/blob/f1f10dd7042126664c10bc1a06773b3422a19b44/msat/src/msat_term.cpp#L124

I also added a unit test using the `googletest` infrastructure. It took a bit to get the parametrized tests right, but it seems to be working now. I added `available_solvers.{h,cpp}` to be used by the tests so that they can determine which solvers to run tests for. I'm open to cleaner solutions though. I decided not to make `available_solvers` part of the regular `smt-switch` library because it would require linking all the solver backends at once. Right now, it's a feature that the different backends are independent and thus you only need to link with the solver backend you're currently using.


tests/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+if (BUILD_MSAT)
+  add_definitions(-DBUILD_MSAT)
+endif(BUILD_MSAT)
+
+add_library(available-solvers "${SMT_SWITCH_LIB_TYPE}"
+  "${PROJECT_SOURCE_DIR}/tests/available_solvers.cpp")
+
+target_include_directories(available-solvers PUBLIC "${PROJECT_SOURCE_DIR}/tests/")
+# TODO: Fix this
+target_link_libraries(available-solvers smt-switch ${EXTRA_LIBS})
```
amaleewilson: What is this TODO and what is EXTRA_LIBS?

makaimann: Oops, good catch! I addressed it with 5369eff91efb3bd01fa72f5d06c2f364591c0b34. The todo was just left over, and I gave EXTRA_LIBS a better name. It's the list of solver backend libraries that are needed.

makaimann:
Going to merge this into the python branch -- it's related to it and I need it there to add more tests.

PR #39, opened 2020-02-17
Add the ability to build python wheels

Kuree:
Changes:
- Add improved find python in cmake
- Add build wheel script in `contrib/wheels`

TODO:
- Use `manylinux2010` docker image to build and repair wheels so that they can be uploaded to PyPi. Given the number of Python versions to support and the amount of time to build `cvc4`, I'd recommend to use Github Actions, where you can have up to 40 concurrent jobs. This will be submitted as a separate PR.

What to link and publish to PyPi:
- In this PR I only build `btor` as a dependency. I believe that we can link `btor` and `cvc4` and distribute the wheels via PyPi. This should be allowed by BSD licence, as long as the copyright info is displayed properly. However, I'm not sure about `msat`.


contrib/wheels/build_wheel.py
```diff
@@ -0,0 +1,118 @@
...
+        return tag == "linux"
+
+    def build_extension(self, ext):
+        extdir = os.path.abspath(
+            os.path.dirname(self.get_ext_fullpath(ext.name)))
+        if not os.path.isdir(extdir):
+            os.mkdir(extdir)
+
+        # test env variable to determine whether to build in debug
+        if os.environ.get("DEBUG") is not None:
```
makaimann: Should this be environment variable be set somewhere? The configure script sets the `CMAKE_BUILD_TYPE` directly now.

contrib/wheels/build_wheel.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # call configure
+        # default install everything?
+        solvers = ["btor"] #, "cvc4", "msat"]
+        solver_path = {"btor": "boolector", "cvc4": "CVC4", "msat": "mathsat"}
+        root_path = os.path.dirname(os.path.dirname(os.path.abspath(os.path.dirname(__file__))))
+        build_dir = os.path.join(root_path, "build")
+
+        for solver in solvers:
+            solver_dir = os.path.join(root_path, "deps", solver_path[solver])
```
makaimann: The configure script allows specifying different solver locations. I don't think that will work for this script but that's probably okay. This only needs to be run occasionally by a developer, right?

Kuree: Correct. This script is not intended to be run regularly during development process. It is written for CI to build wheels.

contrib/wheels/build_wheel.py
Kuree: I will remove this for now. This script is adapted from the template I have for building wheels. I don't think it's useful to build debug mode wheels here since we already has other methods to do so.

PR #40, opened 2020-02-20
Build and upload wheels on tagged release

Kuree:
This PR uses Github Actions to build and deploy wheels to PyPI. It will only trigger the build if it is a tagged. You can see the test package here: https://pypi.org/project/smt-switch/

To test out, simply do
```
pip3 install smt-swittch
```
On any linux machine. It supports Python 3.5+.

@makaimann:
For now it uses my PyPI credential as part of GitHub secrets. I can change the user name to your PyPI account. Once this gets merged, you need to put your PyPI password in as a secret in Github project settings.

I will also either delete the PyPI package or transfer the ownership to you.

EDIT:
I don't have any physical access to OSX, but I've built wheels using CI before. I will submit a PR later that build wheels on OSX platform. If you're interested, I had some luck with Windows wheels via AppVeyor, so I can try that out as well.

### DIFF
--- a/contrib/wheels/ci.sh
+++ b/contrib/wheels/ci.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+cd smt-switch/
+mkdir -p build
+pip install Cython==0.29 --install-option="--no-cython-compile"
+python contrib/wheels/build_wheel.py bdist_wheel
+auditwheel show dist/*
+auditwheel repair dist/*
--- a/contrib/wheels/ci.sh
+++ b/contrib/wheels/ci.sh
@@ -3,7 +3,7 @@ set -e
 
 cd smt-switch/
 mkdir -p build
-pip install Cython==0.29 --install-option="--no-cython-compile"
+pip install Cython==0.29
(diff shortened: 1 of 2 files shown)


contrib/wheels/ci.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+cd smt-switch/
+mkdir -p build
+pip install Cython==0.29 --install-option="--no-cython-compile"
```
makaimann: Lets actually not use the `--no-cython-compile` option. It speeds up the build but then the bindings are less performant. For the release it should be more optimized.

makaimann:
Just added the GitHub Secret. My PyPi username is `makaim`.

Kuree:
Thanks! Can you do a tagged release to see if it works?

Usually the tag is `v + [version number]`, where the `[version number]` matches with the one in `build_wheel.py`.

makaimann:
> Thanks! Can you do a tagged release to see if it works?

Just did a release and it looks like it worked (failed the first time for some unknown reason so I just reran the jobs. Thanks again!

PR #42, opened 2020-02-24
Simple identity walker

makaimann:
Adds a simple identity walker class that can be inherited and modified for arbitrary term traversal/rebuilding.

As part of this, I made a minor change to the term representation. Previously there was an Op called `Const_Array` that was mostly unused. This is because in smt-switch a constant array is considered a value. Having `Const_Array` as the Op was confusing. Instead, the code now tries to maintain an invariant that if `term.is_value()` returns true, then `term.get_op()` returns a null Op (because values don't have ops).


amaleewilson (approved):
Looks good. Thanks for adding me as a reviewer!

PR #45, opened 2020-03-06
Fixes from pysmt tests

makaimann:
Various bug fixes from running @cdonovick's pysmt backend for smt-switch through the pysmt tests.


btor/src/boolector_solver.cpp
```diff
@@ -700,7 +722,7 @@ Term BoolectorSolver::apply_prim_op(PrimOp op, Term t0, Term t1, Term t2) const
 
       std::shared_ptr<BoolectorTerm> bt0 =
           std::static_pointer_cast<BoolectorTerm>(t0);
-      result = boolector_apply(btor, &args[0], 1, bt0->node);
+      result = boolector_apply(btor, &args[0], 2, bt0->node);
```
amaleewilson: I'm assuming this is correct, but it's not immediately clear to me why this 1 was changed to a 2.

include/term.h
```diff
@@ -72,6 +73,13 @@ class TermIterBase
 class TermIter
 {
  public:
+  // typedefs for marking as an input iterator
+  typedef Term value_type;
+  typedef std::ptrdiff_t difference_type;
+  typedef Term * pointer;
```
amaleewilson: A bonus comment here for what pointer and reference are doing would be nice.

btor/src/boolector_solver.cpp
makaimann: Ahh yeah, good question. It is correct -- fixing a copy-and-paste bug from before. This is telling boolector how many arguments are in args and this particular case should be applying a function to 2 arguments.

include/term.h
makaimann: Good idea, thanks! Gave a link to iterator_traits and what it's used for in 70f04dda050865c2182eaeca6d22ad60013d4f2e. Didn't really explain it in detail, but I think it makes sense to accept it as c++ magic

amaleewilson (commented):
Looks good to me

include/term.h
amaleewilson: It makes more sense now! I hadn't read the inputiterator docs.

makaimann:
Thanks for the review!

PR #48, opened 2020-03-17
Remove C++17 Support

makaimann:
Remove anything that required C++17 and instead rely only on C++11.

### DIFF
--- a/btor/src/boolector_factory.cpp
+++ b/btor/src/boolector_factory.cpp
@@ -7,7 +7,7 @@ namespace smt {
 /* BoolectorSolverFactory implementation */
 SmtSolver BoolectorSolverFactory::create()
 {
-  return std::make_unique<BoolectorSolver>();
+  return std::make_shared<BoolectorSolver>();
 }
 /* end BoolectorSolverFactory implementation */
 
--- a/include/utils.h
+++ b/include/utils.h
@@ -18,18 +18,15 @@ const std::size_t global_log_level = _LOGGING_LEVEL;
 const std::size_t global_log_level = 0;
 #endif
 
+// TODO: Create an Assert with an optional message argument
 inline void Assert(bool assertion)
 {
-#if defined(_ASSERTIONS)
   assert(assertion);
-#endif
 }
 
 inline void Unreachable()
 {
-#if defined(_ASSERTIONS)
-    assert(false);
-#endif
+  assert(false);
 }
 
 // logs to stdout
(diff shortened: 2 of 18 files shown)


btor/src/boolector_factory.cpp
```diff
@@ -7,7 +7,7 @@ namespace smt {
 /* BoolectorSolverFactory implementation */
 SmtSolver BoolectorSolverFactory::create()
 {
-  return std::make_unique<BoolectorSolver>();
+  return std::make_shared<BoolectorSolver>();
```
hannes-steffenhagen-diffblue: 👎 There doesn't seem to be an obvious reason for changing this to `shared_ptr`; `unique_ptr` is in C++11, just `make_unique` isn't and you could fix that by simply adding something like

```
template<typename T, typename ...Args>
std::unique_ptr<T> smt_switch_make_unique(Args&&... args) {
  return std::unique_ptr<T>{new T(std::forward<Args>(args)...)};
}
```

include/utils.h
```diff
@@ -5,44 +5,38 @@
 #include "assert.h"
 
 #ifdef _DEBUG
-constexpr bool debug_mode = true;
```
hannes-steffenhagen-diffblue: 👎 `constexpr` is available in C++11.

include/utils.h
```diff
@@ -5,44 +5,38 @@
...
+const std::size_t global_log_level = 0;
 #endif
 
 inline void Assert(bool assertion)
 {
-  if constexpr (assertion_mode)
-  {
-    assert(assertion);
-  }
+#if defined(_ASSERTIONS)
```
hannes-steffenhagen-diffblue: :-1 There is no point in having this flag, `assert` is already a macro that doesn't do anything if `NDEBUG` is set. If you want this to be a more SMT-switch specific thing its name should indicate that somehow.

btor/src/boolector_factory.cpp
makaimann: Thanks for the feedback: it's being switched to `make_shared` because `SmtSolver` is a `shared_ptr` not `unique_ptr` in master now.

include/utils.h
makaimann: Yeah this is currently unused entirely. Eventually it could be a more complicated macro, but for now you're right that it's unneeded.

include/utils.h
makaimann: Good point, reverted with 030be213477218892a8442cf622e80cf71d962be

makaimann: Just fyi constexpr if is not available until C++ 17: https://en.cppreference.com/w/cpp/language/if. Thus there's no point in making the bool a constexpr here.

PR #51, opened 2020-03-19
Implement dump_smt2 for MathSAT backend

makaimann:


msat/src/msat_solver.cpp
```diff
@@ -846,7 +847,22 @@ Term MsatSolver::substitute(const Term term,
 
 void MsatSolver::dump_smt2(FILE * file) const
 {
-  throw NotImplementedException("Can't dump all assertions to a file yet");
+  size_t num_asserted;
+  msat_term * assertions = msat_get_asserted_formulas(env, &num_asserted);
+  msat_term * ait = assertions;
```
ahmed-irfan: why do we need this extra pointer

msat/src/msat_solver.cpp
```diff
@@ -846,7 +847,22 @@ Term MsatSolver::substitute(const Term term,
...
+  msat_term all_asserts = msat_make_true(env);
+  for (size_t i = 0; i < num_asserted; i++)
```
ahmed-irfan: ++i

msat/src/msat_solver.cpp
```diff
@@ -846,7 +847,22 @@ Term MsatSolver::substitute(const Term term,
...
+  {
+    all_asserts = msat_make_and(env, all_asserts, *ait);
+    ait++;
```
ahmed-irfan: ++ait

msat/src/msat_solver.cpp
makaimann: Oh yeah, good catch. That was left over from something else I was trying. Fixed with 75bb8e4bc5d592bc446f0f4debfa68caf44ada46

PR #55, opened 2020-03-24
Add logging infrastructure to store a shadow DAG at smt-switch level

makaimann:
Not all solvers behave exactly the same for term manipulation. In particular, many solvers alias sorts (e.g. `Bool` and `(_ BitVec 1)` or `Int` and `Real`). It's also quite common for solvers to rewrite on the fly, which can make handling terms and/or transferring terms to other solvers complicated.

This PR would add an optional wrapper solver/term/sort that inherits from the same abstract classes, but also keeps a specific solver implementation as a member. For example, `LoggingSolver` is instantiated with an arbitrary `SmtSolver`. Note: nesting `LoggingSolver`s might work but is untested and not encouraged.

Both `LoggingSolver` and `LoggingSort` are pretty straightforward. `LoggingTerm` on the other hand is a bit more interesting. It stores the underlying term, but also the Op and children used to create the term. To avoid exponential blow-up by representing a DAG as a tree, it performs hash-consing using a very simple HashTable implemented with `std::unordered_map` and `smt::UnorderedTermSet`. Importantly, it relies on the underlying term for some features, and implements others itself. 

For example, it uses the hash from the underlying term, but implements comparison. This is useful for solvers that alias sorts or rewrite on the fly. 

Example 1: If `Bool` and `(_ BitVec 1)` are equivalent in the underlying solver, then `true` and `#b1` are equivalent. However `LoggingTerm` checks the sorts (which are `LoggingSort`s) and returns false for the comparison.

Note: What about terms that the solver takes initiative in creating? We make the assumption that the solver only generates certain classes of terms. For example, values. If the solver generates a value, this is typically prompted by `get_value(Term & t)`. In this case, we can check the sort. Thus, if we ask for the value of a boolean symbolic constant, `b`, and the solver gives `#b1`, we know that the sort is supposed to be `Bool` so it will set the sort accordingly.

Example 2: Rewriting on the fly. A solver might rewrite the term `(= x x)` to `true`. However, arbitrary rewritings can cause problems for the logging infrastructure. Thus, we also use the Op and children for `LoggingTerm` comparison. Thus, even if the underlying solver considers `(= x x)` and `true` to be equivalent, at the logging smt-switch level, they are not.

I'd like to call attention to a couple areas that I think are especially in need of feedback:
* does the `LoggingTerm::compare` implementation seem reasonable?
* the `LoggingTerm` currently builds it's own `to_string` representation on demand. Does this seem reasonable? I chose to do it lazily because if people aren't printing terms it doesn't seem worth it to store large strings.
  * additionally, the `to_string` relies on the underlying term for values, because those might have come from the solver. Does that make sense?
* which solvers should it wrap?
  * currently it only wraps Boolector, because of the sort aliasing.
  * MathSAT is pretty smt-lib compliant in terms of sorts and term representation. However, it does rewrite on the fly. should we wrap it?
  * CVC4 aliases `Int` and `Real`. Should we wrap it?
* I added a specific function for getting array values (`get_array_values`). This is because it was difficult to wrap the underlying terms if we returned non-value terms (e.g. nested stores). Does this make sense? It's not a standard SMT-LIB feature anyway.
  * CVC4 can return array models with `get_value` currently. Should I disable this in smt-switch so that it behaves the same as other solvers? E.g. throws an exception on `s->get_value(arr_var)`?
* Boolector uses a `LoggingSolver` wrapper by default now
  * but it also allows creating a "raw" solver with `create_lite_solver`. Do you like this name? Also, is it confusing that `create` sometimes creates a LoggingSolver and sometimes a regular one? I just want the default to be the best configuration. E.g. CVC4 might not need a wrapper ever.


makaimann:
This PR is ready to be reviewed imo, but please don't merge it yet. I'd like to get a couple other PRs merged first, and this one might require minor changes to align it with master afterwards. For example, once the Yices backend is merged, I'll update it to use a `LoggingSolver` wrapper.

makaimann:
All right, I've updated it for Yices2 as well, so I think it's ready to be merged as soon as it passes review.

PR #60, opened 2020-05-04
Add unsat core support

makaimann:
Adds support for querying backend solvers for unsat cores. After some discussion, we decided the interface should use `check_sat_assuming`. Thus, after a call to `check_sat_assuming`, `get_unsat_core` will return a vector with a subset of the assumptions that are sufficient to make the assertions unsat.

### DIFF
--- a/btor/src/boolector_solver.cpp
+++ b/btor/src/boolector_solver.cpp
@@ -86,6 +86,13 @@ void BoolectorSolver::set_opt(const std::string option, const std::string value)
       boolector_set_opt(btor, BTOR_OPT_INCREMENTAL, 1);
     }
   }
+  else if (option == "produce-unsat-cores")
+  {
+    produce_unsat_cores = true;
+    unsat_core_assumptions.push_back(std::vector<BoolectorNode *>());
+    // needs to be incremental
+    boolector_set_opt(btor, BTOR_OPT_INCREMENTAL, 1);
+  }
   else
   {
     std::string msg("Option ");
@@ -200,11 +207,31 @@ void BoolectorSolver::assert_formula(const Term & t)
 {
   std::shared_ptr<BoolectorTerm> bt =
       std::static_pointer_cast<BoolectorTerm>(t);
-  boolector_assert(btor, bt->node);
+  if (produce_unsat_cores)
+  {
+    // unsat core production needs to use assumptions
+    unsat_core_assumptions[unsat_core_assumptions.size() - 1].push_back(
+        bt->node);
+  }
+  else
+  {
+    boolector_assert(btor, bt->node);
+  }
 }
 
 Result BoolectorSolver::check_sat()
 {
+  if (produce_unsat_cores)
+  {
+    for (auto assumptions : unsat_core_assumptions)
+    {
+      for (auto a : assumptions)
+      {
+        boolector_assume(btor, a);
+      }
+    }
+  }
+
   if (boolector_sat(btor) == BOOLECTOR_SAT)
   {
     return Result(SAT);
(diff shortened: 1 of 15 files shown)


makaimann:
@ahmed-irfan Does this seem like the right way to implement unsat cores for mathsat? It's a bit complicated but then it matches CVC4 where the unsat core is over all assertions (for mathsat/boolector we keep track of assumptions at the smt-switch level). Alternatively, we could make an interface where you need to "tag" certain assertions for them to be added to the unsat core and then only return those. For CVC4, we could use `getUnsatAssumptions`. What do you think?

ahmed-irfan:
Thanks Makai for working on this feature. i will check it soon

btor/src/boolector_solver.cpp
```diff
@@ -200,11 +210,31 @@ void BoolectorSolver::assert_formula(const Term & t)
 {
   std::shared_ptr<BoolectorTerm> bt =
       std::static_pointer_cast<BoolectorTerm>(t);
-  boolector_assert(btor, bt->node);
+  if (produce_unsat_cores)
+  {
+    // unsat core production needs to use assumptions
+    unsat_core_assumptions[unsat_core_assumptions.size() - 1].push_back(
```
ahmed-irfan, line 216: i don't understand this

makaimann: That isn't clear, you're right. Added a comment with: fa8376eef19d36053e7105c6a0fa93d5bed5e721

ahmed-irfan: OK. what is the assumption here?

makaimann: Right, so there's an implicit assumption that `unsat_core_assumptions` has one more elements than the current context level. And then, we just need to use `unsat_core_assumptions` because Boolector gets unsat cores using assumptions and those get removed after each call to `check_sat` so we need to maintain a list of the assertions to add before checking sat. This is one of the reasons I wanted you to review this actually. Do you think there's a better way to implement this when it depends on assumptions? With this implementation, we could lose performance because of solving with assumptions instead of assertions.

msat/src/msat_solver.cpp
```diff
@@ -900,6 +934,17 @@ void MsatSolver::dump_smt2(std::string filename) const
   fclose(file);
 }
 
+// helpers
+msat_term MsatSolver::label(msat_term p) const
+{
+  std::ostringstream buf;
+  buf << ".ref_axiom_lbl{" << msat_term_id(p) << "}";
```
ahmed-irfan, line 941: maybe change into `assump_lbl{`

btor/src/boolector_solver.cpp
```diff
@@ -348,6 +362,20 @@ Term BoolectorSolver::get_value(Term & t) const
...
+TermVec BoolectorSolver::get_unsat_core()
+{
+  TermVec core;
+  BoolectorNode ** bcore = boolector_get_failed_assumptions(btor);
+  BoolectorNode ** bcore_iter = bcore;
+  while (*bcore_iter)
+  {
+    core.push_back(std::make_shared<BoolectorTerm>(
+        btor, boolector_copy(btor, *bcore_iter)));
+    bcore_iter++;
```
ahmed-irfan, line 374: minor: change this to `++bcore_iter;`

msat/src/msat_solver.cpp
```diff
@@ -252,6 +264,28 @@ Term MsatSolver::get_value(Term & t) const
...
+    throw InternalSolverException(
+        "Got an empty unsat core. Ensure your last call was unsat and had "
+        "assumptions in check_sat_assuming that are required to get an unsat "
+        "result");
+  }
+  msat_term * mcore_iter = mcore;
+  for (size_t i = 0; i < core_size; ++i)
+  {
+    core.push_back(std::make_shared<MsatTerm>(env, *mcore_iter));
+    mcore_iter++;
```
ahmed-irfan, line 283: minor: change it to `++mcore_iter;`

msat/src/msat_solver.cpp
makaimann: Good catch! Changed with b3b449b9de26923782e031258060f4133a681113

btor/src/boolector_solver.cpp
makaimann: I actually simplified that whole function even more. It doesn't need a separate iterator, because that array doesn't need to be freed (Boolector manages the memory of the core). 13332aa639fd719a67f459b665a16a240d47a984

msat/src/msat_solver.cpp
makaimann: Done 311fd024b17bc15188dc5908e971282911afa01a

makaimann:
Thanks for the review!

PR #65, opened 2020-06-02
some utils methods

ahmed-irfan:
adds conjunctive and disjunctive paritition methods

adds get_free_vars method

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -20,6 +20,9 @@
 #include <iostream>
 #include "assert.h"
 
+#include "smt.h"
+#include "ops.h"
+
 #ifndef NDEBUG
 #define _ASSERTIONS
 #endif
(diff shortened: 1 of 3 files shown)


include/utils.h
```diff
@@ -49,4 +51,11 @@ inline void Log(std::string msg)
   }
 }
 
+// term helper methods
+void conjunctive_partition(const smt::Term &term, smt::TermVec &out);
+
+void disjunctive_partition(const smt::Term &term, smt::TermVec &out);
+
+void get_free_vars(const smt::Term &term, smt::TermVec &out);
```
makaimann, line 59: Let's call this `get_free_symbols` or `get_free_symbolic_consts`. I've been staying away from `vars` to avoid the var/const confusion. I am planning to make a change soon where there's a distinction between `is_symbol` and `is_symbolic_const` where the former includes uninterpreted functions. So you can make the name match whichever functionality you expect, i.e. `get_free_symbols` if it will include uninterpreted functions and otherwise `get_free_symbolic_consts`.

ahmed-irfan: makes sense. I changed the function name accordingly.

PR #69, opened 2020-06-11
Printing solver interface

yoni206:
This PR adds the interface to create and use a "printing solver", that is, a solver that dumps each SMT-LIB command it performs to a stream.
The current PR only adds the .h file, an almost empty cpp file, and the required change to cmake. It also includes documentation.
A subsequent PR will add all implementations and tests.

### DIFF
--- a/include/printing_solver.h
+++ b/include/printing_solver.h
@@ -49,6 +49,7 @@ class PrintingSolver : public AbsSmtSolver
 
   /* Operators that are printed */
   Sort make_sort(const std::string name, uint64_t arity) const override;
+  Term make_symbol(const std::string name, const Sort & sort) override;
 
   Term get_value(const Term & t) const override;
   UnorderedTermMap get_array_values(const Term & arr,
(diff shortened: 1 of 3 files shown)


include/printing_solver.h
```diff
@@ -0,0 +1,131 @@
...
+  void push(uint64_t num = 1) override;
+  void pop(uint64_t num = 1) override;
+  void reset_assertions() override;
+  bool get_interpolant(const Term & A,
+                               const Term & B,
+                               Term & out_I) const override;
+
+
+  /* Operators that are not printed 
+   * For example, creating terms is not printted, but the
```
makaimann, line 72: minor typo on "printed"

include/printing_solver.h
```diff
@@ -0,0 +1,131 @@
...
+  Term get_tester(const Sort & s, std::string name) const override;
+  Term get_selector(const Sort & s, std::string con, std::string name) const override;
+
+  Term make_term(bool b) const override;
+  Term make_term(int64_t i, const Sort & sort) const override;
+  Term make_term(const std::string val,
+                 const Sort & sort,
+                 uint64_t base = 10) const override;
+  Term make_term(const Term & val, const Sort & sort) const override;
+  Term make_symbol(const std::string name, const Sort & sort) override;
```
makaimann, line 102: Shouldn't this one be printed? You would need the `declare-fun` right?

yoni206:
Thanks for the review. Addressed in https://github.com/makaimann/smt-switch/pull/69/commits/c44560d58177d3562a7bd61af16d872ae111926f.

PR #70, opened 2020-06-15
Printing solver: implementations

yoni206:
This PR adds implementations and tests for the interface decalred in #69 .


include/printing_solver.h
```diff
@@ -30,11 +30,11 @@ namespace smt {
  */
 enum PrintingStyleEnum
 {
-  DEFAULT = 0,
-  BTOR_FMT,
-  CVC4_FMT,
-  MSAT_FMT,
-  YICES_FMT2
+  DEFAULT_STYLE = 0,
+  BTOR_STYLE,
```
makaimann, line 34: Quick question on this: do you actually use all these different styles? I'm wondering if we should only include the relevant ones. Otherwise users might expect different output if they pass different styles. The `BTOR_STYLE` is probably unused, right?

src/printing_solver.cpp
```diff
@@ -11,6 +11,8 @@
 **
 ** \brief Class that wraps another SmtSolver and dumps SMT-LIB
 **        that corresponds to the operations being performed.
+**        The SMT-LIB sommand is printing before being executed,
```
makaimann, line 14: Useful description, thanks! Minor suggestion: printing -> printed

src/printing_solver.cpp
```diff
@@ -11,6 +11,8 @@
...
+**        So that in case of an error it easier to see when it happened.
```
makaimann, line 15: minor suggestion: lower-case "s"

src/printing_solver.cpp
```diff
@@ -40,5 +42,247 @@ namespace smt {
 /* PrintingSolver */
 
 // implementations
+PrintingSolver::PrintingSolver(SmtSolver s, std::ostream* os, PrintingStyleEnum pse)
+    : AbsSmtSolver(s->get_solver_enum()),
```
makaimann, line 46: What do you think. Should a printing solver get its own enum? I'm not sure if it needs one (i.e. I think it's fine how you've done it). But what do you think? Would the user expect a different solver enum for a CVC4Solver vs a PrintingSolver that wraps a CVC4Solver?

tests/msat/msat-printing.cpp
```diff
@@ -0,0 +1,81 @@
+/*********************                                                        */
+/*! \file cvc4-printinh.cpp
```
makaimann, line 2: filename

tests/msat/msat-printing.cpp
```diff
@@ -0,0 +1,91 @@
...
+  // STRFY is defined in test-utils.h and converts
+  // a macro to its string representation
+  string command(STRFY(MSAT_HOME));
+  command += "/bin/mathsat ";
+  command += filename;
+  std::cout << "Running command: " << command << std::endl;
+  string result = exec(command.c_str());
+  std::cout << "got result:\n" << result << std::endl;
+  assert(result == "unsat\n(<= 2 (+ z (* (- 1) x)))\n");
+  return 0;
```
makaimann, line 90: what do you think about cleaning up the generated file if the test passes? I think it's nice to not have test files left around.

tests/cvc4/cvc4-printing.cpp
```diff
@@ -0,0 +1,111 @@
...
+  // that points to the top-level CVC4 directory
+  // STRFY is defined in test-utils.h and converts
+  // a macro to its string representation
+  string command(STRFY(CVC4_HOME));
+  command += "/build/bin/cvc4 ";
+  command += filename;
+  std::cout << "Running command: " << command << std::endl;
+  string result = exec(command.c_str());
+  std::cout << "got result:\n" << result << std::endl;
+  assert(result == "unsat\n()\nsat\n((x (_ bv0 32)))\n");
```
makaimann, line 109: what do you think about cleaning up the generated file if the test passes? I think it's nice to not have test files left around.

src/printing_solver.cpp
```diff
@@ -40,5 +42,247 @@ namespace smt {
...
+   * The printing follows the internal implementation from msat_solver.h
+   * in which the assertions are labeled by interpolation groups
+   */
+  assert(style == PrintingStyleEnum::MSAT_STYLE);
+  (*out_stream) << "(" << ASSERT_STR << " (! " << A << " :" << INTERPOLATION_GROUP_STR << " g1))" << endl;
+  (*out_stream) << "(" << ASSERT_STR << " (! " << B << " :" << INTERPOLATION_GROUP_STR << " g2))" << endl;;
+  (*out_stream) << "(" << CHECK_SAT_STR << ")" << endl;
+  (*out_stream) << "(" << MSAT_GET_INTERPOLANT_STR << " (g1)" << ")" << endl;
+  (*out_stream) << "; when running mathsat, use `-interpolation=true` flag" << endl;
+  wrapped_solver->get_interpolant(A, B, out_I);
```
makaimann, line 280:
```suggestion
  return wrapped_solver->get_interpolant(A, B, out_I);
```

src/printing_solver.cpp
yoni206: yeah i wasn't sure about that either... My current code is a bit guilty in misrepresenting the actual state, but having more enums for printing solvers seems bloated... I'll leave the decision to you :)

makaimann: Yeah it would be a lot of enums. Let's leave it as-is for now! If it becomes important, we can change it later.

makaimann:
Just FYI, I had to change this line https://github.com/yoni206/smt-switch/blob/8790c4edb030bbef00673885de23aab922c88605/tests/cvc4/cvc4-printing.cpp#L90. Originally it was in a C assert, but then the test fails if you don't compile with debug. However, this made me realize that Travis wasn't building debug so it's good this came up!

I'm just waiting for the Travis run to complete and then I can merge this.

makaimann:
Looks like the MathSAT test fails on Mac: https://travis-ci.org/github/makaimann/smt-switch/jobs/699491315.

I'll try to debug it on Travis, but if you're able to reproduce it locally, that might be easier.

yoni206:
woooohoooo!

PR #71, opened 2020-06-17
Add support for quantifiers

makaimann:
This PR would add support for quantifiers. The implementations are primarily straightforward. The interface changes include `Forall` and `Exists` `PrimOps`, as well as a `make_param` and corresponding `is_param` functions for creating and identifying parameters (variables) that will be bound by a quantifier.

One complication is related to term traversal. It seemed overly complicated and difficult to support a list of bound variables due to the different implementations in various solvers. Instead, I restrict quantifiers in `smt-switch` to bind only one parameter at a time. Of course, this has no limitation on expressivity because you can always nest quantified statements.

@barrettcw You don't need to review the whole PR, but I'd appreciate your feedback on the above paragraph if you have time.

@ahmed-irfan If you don't mind reviewing, I'd appreciate it. And thanks for doing lots of reviews lately!

Notice that the changes to `is_symbolic_const` might affect lazybv2int. So now, there are three functions:
* `is_symbol` - returns true for all functions, parameters and symbolic constants
* `is_symbolic_const` - returns true for uninterpreted, symbolic constants (not functions or parameters)
* `is_param` - returns true for parameters (i.e. bound variables)

Prior to this change, `is_symbolic_const` actually returned true for functions as well, so lazybv2int might want to switch calls to `is_symbolic_const` to `is_symbol`.

One last thing to bring up. In the first quantifier test, mathsat returns unknown. This seems strange because it's a very simple query. Do you happen to know if there are any flags I should set? Or maybe quantification over booleans is not supported in mathsat?


makaimann:
Just FYI this is reporting the travis build as failed but that's just because Travis non-deterministically fails sometimes. I restarted it and it passed: https://travis-ci.org/github/makaimann/smt-switch/builds/699155124

ahmed-irfan:
sure, I can review this.

cvc4/src/cvc4_solver.cpp
```diff
@@ -99,6 +99,8 @@ const std::unordered_map<PrimOp, ::CVC4::api::Kind> primop2kind(
       { Int_To_BV, ::CVC4::api::INT_TO_BITVECTOR },
       { Select, ::CVC4::api::SELECT },
       { Store, ::CVC4::api::STORE },
+      { Forall, ::CVC4::api::FORALL },
+      { Store, ::CVC4::api::EXISTS },
```
yoni206, line 103:
```suggestion
      { Exists, ::CVC4::api::EXISTS },
```

cvc4/src/cvc4_solver.cpp
```diff
@@ -827,6 +856,12 @@ Term CVC4Solver::make_term(Op op,
   }
   catch (::CVC4::api::CVC4ApiException & e)
   {
+    if (op.prim_op == Forall || op.prim_op == Exists)
+    {
+      throw IncorrectUsageException(
```
yoni206, line 861: Why is that? The standard allows multiple.

btor/src/boolector_solver.cpp
```diff
@@ -624,8 +634,22 @@ Term BoolectorSolver::make_symbol(const std::string name, const Sort & sort)
   return term;
 }
 
+Term BoolectorSolver::make_param(const std::string name, const Sort & sort)
+{
+  std::shared_ptr<BoolectorSortBase> bs =
+      std::static_pointer_cast<BoolectorSortBase>(sort);
+  BoolectorNode * n = boolector_param(btor, bs->sort, name.c_str());
+  Term term = std::make_shared<BoolectorTerm>(btor, n);
```
ahmed-irfan, line 642: maybe return directly

btor/src/boolector_solver.cpp
```diff
@@ -624,8 +634,22 @@ Term BoolectorSolver::make_symbol(const std::string name, const Sort & sort)
...
+  return term;
+}
+
 Term BoolectorSolver::make_term(Op op, const Term & t) const
 {
+  if (op.prim_op == Forall || op.prim_op == Exists)
```
ahmed-irfan, line 648: i don't understand this check. Can you explain what you are doing here?

cvc4/src/cvc4_solver.cpp
makaimann: Oops, good catch. Thanks!

cvc4/src/cvc4_solver.cpp
makaimann: Because it got really complicated to support term traversal if I allowed more than one. CVC4 uses a `BOUND_VAR_LIST` that would have to be flattened. Or if we had a smt-switch version of `BOUND_VAR_LIST` then we'd have to transform all the other solvers to this representation. I'm open to other ideas, but this was the simplest thing I could think of. Then I just always assume that the term is `quantifier param . term`. And it's still equally expressive, because you can always nest them.

btor/src/boolector_term.cpp
```diff
@@ -267,9 +265,27 @@ Sort BoolectorTerm::get_sort() const
   return sort;
 }
 
+bool BoolectorTerm::is_symbol() const
+{
+  // functions, parameters, and symbolic constants are all symbols
+  auto bkind = bn->kind;
+  return !negated
```
ahmed-irfan, line 272: a side note: we should have different naming convention for data members.

ahmed-irfan: i am referring to `negated` here

btor/src/boolector_term.cpp
```diff
@@ -267,9 +265,27 @@ Sort BoolectorTerm::get_sort() const
...
+  return !negated && (bn->kind == BTOR_PARAM_NODE);
+}
+
 bool BoolectorTerm::is_symbolic_const() const
 {
-  return is_sym;
+  bool is_fun = boolector_is_fun(btor, node);
+  auto bkind = bn->kind;
+  bool is_sym =
+      !negated && ((bkind == BTOR_VAR_NODE) || (bkind == BTOR_UF_NODE));
```
ahmed-irfan, line 287: are we sure that we want to check for uf-node?

btor/src/boolector_solver.cpp
makaimann: Good idea -- done with 32ad75043d8ea7fd7946d04c3ffdc5de740395e2

cvc4/src/cvc4_solver.cpp
```diff
@@ -617,6 +619,21 @@ Term CVC4Solver::make_symbol(const std::string name, const Sort & sort)
   }
 }
 
+Term CVC4Solver::make_param(const std::string name, const Sort & sort)
+{
+  try
+  {
+    std::shared_ptr<CVC4Sort> csort = std::static_pointer_cast<CVC4Sort>(sort);
+    ::CVC4::api::Term t = solver.mkVar(csort->sort, name);
+    Term res = std::make_shared<::smt::CVC4Term>(t);
```
ahmed-irfan, line 628: return directly

btor/src/boolector_solver.cpp
makaimann: It's not a valid use of `make_term` with those operators, because it needs to take exactly two terms: the parameter to be bound, and the body term. Since that function only takes an `op` and one `term`, it can't be a valid use. So the exception is saying, you need to use a parameter and a body term with a quantifier op. Does that make sense?

btor/src/boolector_term.cpp
makaimann: Ahh like `negated_`? Yeah that's not used anywhere in the codebase, but would probably be a good idea.

btor/src/boolector_term.cpp
makaimann: Yeah so the way I've set it up is that `is_symbol` returns true for any uninterpreted constant, or function. `is_symbolic_const` only returns true for uninterpreted constants, and `is_param` only returns true for parameters. Does that seem reasonable?

cvc4/src/cvc4_solver.cpp
makaimann: Thanks, done with 4560d2269450597426f0b7792322c8082f717d4b

src/logging_solver.cpp
```diff
@@ -476,7 +497,11 @@ void LoggingSolver::assert_formula(const Term & t)
   wrapped_solver->assert_formula(lt->wrapped_term);
 }
 
-Result LoggingSolver::check_sat() { return wrapped_solver->check_sat(); }
+Result LoggingSolver::check_sat()
+{
+  Result r = wrapped_solver->check_sat();
```
ahmed-irfan, line 502: return directly

tests/btor/btor-substitute.cpp
```diff
@@ -74,15 +74,15 @@ int main()
         to_visit.push_back(c);
       }
 
-      if (t->is_symbolic_const())
+      if (t->is_symbol())
       {
-        num_consts++;
+        num_symbols++;
```
ahmed-irfan, line 79: `++num_symbols;`

tests/btor/btor-substitute.cpp
```diff
@@ -107,15 +107,15 @@ int main()
...
```
ahmed-irfan, line 112: `++num_symbols;`

yices2/src/yices2_term.cpp
```diff
@@ -364,8 +364,32 @@ Sort Yices2Term::get_sort() const
...
+}
+
 bool Yices2Term::is_symbolic_const() const
 {
+  // functions are not constants
+  if (is_function)
+  {
+    return false;
+  }
+
```
ahmed-irfan, line 392: `return is_symbol();`

btor/src/boolector_term.cpp
```diff
@@ -267,9 +265,27 @@ Sort BoolectorTerm::get_sort() const
...
+  return is_sym && !is_fun;
```
ahmed-irfan, line 288: maybe `return !is_fun && !is_param && is_symbol();`

ahmed-irfan (commented):
Left few comments. Please take a look when you can.

ahmed-irfan:
For the question regarding MathSAT, AFAIK, the quantifiers support is very limited: you can create quantified terms at the API level or parse quantified SMT problems.

btor/src/boolector_term.cpp
makaimann: Oh I see, yeah this could be simpler. I just copy and pasted. But I could just take out the UF_NODE one and then not have to check if it's a function separately. I'll do that.

cvc4/src/cvc4_term.cpp
```diff
@@ -221,9 +238,22 @@ Sort CVC4Term::get_sort() const
   return std::make_shared<CVC4Sort> (term.getSort());
 }
 
+bool CVC4Term::is_symbol() const
+{
+  // functions, parameters, and symbolic constants are all symbols
+  ::CVC4::api::Kind k = term.getKind();
+  return (k == ::CVC4::api::CONSTANT || k == ::CVC4::api::VARIABLE);
```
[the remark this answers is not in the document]
makaimann: Yep, specifically this is tested here: https://github.com/makaimann/smt-switch/blob/4c4d48371a6b7dd40b23bee34ffd61b7266de79e/tests/unit/unit-term.cpp#L53

btor/src/boolector_term.cpp
makaimann: Made it simpler in e3e74f36278132d722bf3991c6515fa38de16c1b

btor/src/boolector_term.cpp
makaimann: Made it simpler in e3e74f36278132d722bf3991c6515fa38de16c1b.

src/logging_solver.cpp
makaimann: Addressed in e3e74f36278132d722bf3991c6515fa38de16c1b

tests/btor/btor-substitute.cpp
makaimann: Addressed in e3e74f36278132d722bf3991c6515fa38de16c1b. Just out of curiosity, why do you prefer prefix in this case?

tests/btor/btor-substitute.cpp
makaimann: Addressed in e3e74f36278132d722bf3991c6515fa38de16c1b

yices2/src/yices2_term.cpp
makaimann: Addressed in e3e74f36278132d722bf3991c6515fa38de16c1b

ahmed-irfan (approved):
LGTM

makaimann:
> For the question regarding MathSAT, AFAIK, the quantifiers support is very limited: you can create quantified terms at the API level or parse quantified SMT problems.

Gotcha! Do you think I should not tag it as supporting quantifiers then?

makaimann:
@yoni206 you okay with this PR as-is also?

ahmed-irfan:
> > For the question regarding MathSAT, AFAIK, the quantifiers support is very limited: you can create quantified terms at the API level or parse quantified SMT problems.
> 
> Gotcha! Do you think I should not tag it as supporting quantifiers then?

Before answering that, can you check if MathSAT throws some exception if calling check-sat on a quantified problem?

yoni206 (approved):

makaimann:
> > > For the question regarding MathSAT, AFAIK, the quantifiers support is very limited: you can create quantified terms at the API level or parse quantified SMT problems.
> > 
> > 
> > Gotcha! Do you think I should not tag it as supporting quantifiers then?
> 
> Before answering that, can you check if MathSAT throws some exception if calling check-sat on a quantified problem?

There's no exception or anything, it just returns unknown.

makaimann:
I'm gonna merge this now, and if we decide to not mark MathSAT as supporting quantifiers, we can do that in a follow-up PR.

PR #72, opened 2020-06-17
Add option to support reset_assertions in btor backend with contexts

makaimann:
Boolector does not support reset_assertions. We can approximate it at the Smt-switch level by always solving at context 1 and popping back to context 0 to reset assertions (then pushing to context 1 again before continuing).

I decided to have this option disabled by default because it likely impacts performance. But, the exception that is thrown will tell the user what option to set.

### DIFF
--- a/tests/unit/unit-reset-assertions.cpp
+++ b/tests/unit/unit-reset-assertions.cpp
@@ -95,6 +95,8 @@ TEST_P(UnitResetAssertionsTests, ResetAssertions)
   ASSERT_TRUE(r.is_sat());
 }
 
+// only run this test if built with Boolector
+#ifdef BUILD_BTOR
 TEST(UnitBtorResetAssertionsTests, ResetAssertionsException)
 {
   // Boolector backend does not support reset_assertions by default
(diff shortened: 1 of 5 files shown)


tests/unit/unit-reset-assertions.cpp
```diff
@@ -0,0 +1,125 @@
...
+  Term zero = s->make_term(0, bvsort);
+  Term make_unsat = s->make_term(
+      And, s->make_term(Distinct, x, zero), s->make_term(Distinct, y, zero));
+  Term half_max = s->make_term(128, bvsort);
+  make_unsat = s->make_term(And,
+                            make_unsat,
+                            s->make_term(And,
+                                         s->make_term(BVUlt, x, half_max),
+                                         s->make_term(BVUlt, y, half_max)));
+  s->push(2);
```
ahmed-irfan, line 80: why 2?

makaimann: Good question! No good reason -- just that I need to track the current context level in boolector now, and I wanted to test that it did the right thing even for values other than 1.

ahmed-irfan: ah, ok.

PR #74, opened 2020-06-18
Pass original sort instead of transferred sort to value_from_smt2

makaimann:
This allows transferring terms from a solver that doesn't alias sorts to one that does alias sorts. If we use the transferred sort, then we already lose that information.


makaimann:
Closing because incorporated into https://github.com/makaimann/smt-switch/pull/81

PR #78, opened 2020-07-01
Adds general methods for sort checking and inference

makaimann:
This PR adds functions that can check if an operator applied to some terms is well-sorted. It also moves the sort computation used in the LoggingSolver infrastructure into more general functions (that take a solver as an input). These functions compute the expected sort of applying an operation.

@ahmed-irfan, sorry for sending lots of PRs your way. This is one that is related to the changes we want to make in Pono (passing terms between solvers for different engines). Would you mind reviewing this? I split all the changes into two parts to hopefully make reviewing it easier.


include/sort_inference.h
```diff
@@ -0,0 +1,187 @@
...
+** in the top-level source directory) and their institutional affiliations.
+** All rights reserved.  See the file LICENSE in the top-level source
+** directory for licensing information.\endverbatim
+**
+** \brief Utility functions for checking sortedness and computing the
+**        expected sort when building a term.
+**
+**/
+
+#pragma once
```
ahmed-irfan, line 17: it is not consistent with the rest of the code.
(even though I like the use of pragma once)

makaimann: Good point -- made the change! We could change to pragma once in the future.

PR #81, opened 2020-07-07
2nd part of sort casting PR

makaimann:
This one uses the method from https://github.com/makaimann/smt-switch/pull/78 in the term translator to make sure terms are well-sorted on the fly when translating.


include/boolector_factory.h
```diff
@@ -14,6 +14,8 @@
 **
 **/
 
+#pragma once
```
ahmed-irfan, line 17: I made the PR https://github.com/makaimann/smt-switch/pull/83 to use it in every header file

include/term_translator.h
```diff
@@ -25,17 +25,93 @@
 #include "term.h"
 
 namespace smt {
+
+/** Class for translating terms from *one* other solver to *one* new solver
+ *  will fail if you try to convert terms from more than one solver
```
ahmed-irfan, line 30: can you give an example of this scenario? that would be nice to have.

include/term_translator.h
```diff
@@ -25,17 +25,93 @@
...
+   *  @param sk the expected SortKind of the transferred term
+   *  @return a term belonging to this solver
+   */
+  Term transfer_term(const Term & term, const SortKind sk);
+
   /* Returns reference to cache -- can be used to populate with symbols */
   UnorderedTermMap & get_cache() { return cache; };
 
+  /* Returns a reference to the solver this object translates terms to */
+  SmtSolver & get_solver() { return solver; };
```
ahmed-irfan, line 68: just checking: solver is "intentionaly" mutable here.

src/term_translator.cpp
```diff
@@ -74,82 +113,162 @@ Term TermTranslator::transfer_term(const Term & term)
...
+  {
+    Sort sort = solver->make_sort(INT);
+    return cast_term(transferred_term, sort);
+  }
+  else
+  {
+    string msg("Cannot cast ");
+    msg += transferred_term->to_string() + " to " + smt::to_string(sk);
+    throw IncorrectUsageException(msg);
+  }
```
ahmed-irfan, line 264: what happens if the sort of transferred_sk and sk are the same?

ahmed-irfan (commented):
Left some comments

include/term_translator.h
makaimann: Hmm.. I'm not actually sure how it would be used. Maybe it doesn't need to be mutable. That's a good question -- I'll think on it.

src/term_translator.cpp
makaimann: It will just return the term unaltered: https://github.com/makaimann/smt-switch/blob/5bd71ccdbe37180d79bfdd8598dae1ad2821aa90/src/term_translator.cpp#L230

include/term_translator.h
makaimann: Actually, wait that would just be a const reference to a pointer. So it could still call non-const methods of the solver, it just means it couldn't switch which solver it is. It definitely should be const then. Just made that change.

include/term_translator.h
makaimann: I added an example in the comments with e15b8eef97541ef9a279dca388d0aa58defaea59. Does that seem good?

ahmed-irfan: yeah, thanks

ahmed-irfan (approved):
LGTM. Thanks

makaimann:
Thanks for reviewing!

PR #82, opened 2020-07-08
Logging term compare fix

makaimann:
This is a fix for `LoggingTerm::compare`. This is used to test equality between `LoggingTerms`. Previously, it was checking that the wrapped term, op and sort were the same, and then checking that each of the children were the same. While technically correct, this was terrible for performance because the children are themselves `LoggingTerms`. Thus, equality comparison traversed the whole term DAG.

This PR adds a simple fix based on hash-consing (which was already performed before this PR). Because of hash-consing, we can simply compare the pointers. However, it's not quite that simple because to perform hash-consing we need to compare for equality to find the single pointer that is used for that term in the hash table. Thus, we can't use the pointer of the current term. But, we can use the pointers of the children. So, when checking the children for equality, we compare the pointers instead of using `LoggingTerm::compare`.

This PR also adds a few tests and adds hash table lookups in places they were missing.

### DIFF
--- a/src/logging_term.cpp
+++ b/src/logging_term.cpp
@@ -53,16 +53,18 @@ bool LoggingTerm::compare(const Term & t) const
   }
 
   shared_ptr<LoggingTerm> lt = static_pointer_cast<LoggingTerm>(t);
-  // compare op
-  if (op != lt->op)
+
+  // compare wrapped term and the LoggingSort
+  // this handles values (e.g. null operators and no children)
+  // and because of the sort comparison also handles sort aliasing
+  // of the wrapped solver
+  if (wrapped_term != lt->wrapped_term || sort != lt->sort)
   {
     return false;
   }
 
-  // compare underlying term and sort
-  // this will handle sort aliasing issues from solvers
-  // that don't distinguish between certain sorts
-  if (wrapped_term != lt->wrapped_term || sort != lt->sort)
+  // compare op
+  if (op != lt->op)
   {
     return false;
   }
@@ -77,7 +79,9 @@ bool LoggingTerm::compare(const Term & t) const
   {
     for (size_t i = 0; i < children.size(); i++)
     {
-      if (children[i] != lt->children[i])
+      // because of hash-consing, we can compare the pointers
+      // otherwise would recursively call compare on the LoggingTerm children
+      if (children[i].get() != lt->children[i].get())
       {
         return false;
       }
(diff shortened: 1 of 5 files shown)


src/logging_term.cpp
```diff
@@ -46,6 +46,25 @@ LoggingTerm::~LoggingTerm() {}
...
+  // still need to return false. It needs to check that the sort,
+  // operator and children are the same to keep the contract that
+  // the logging term hides sort aliasing and term rewriting
+  // Furthermore, we want to avoid recursively calling compare
+  // on the whole term DAG because the children are also LoggingTerms
+  // which rely on this method for equality
+  // Because we perform hash-consing, we can only compare the pointer values
+  // to compare children. See the use of .get() on the children below.
+  // However, we do not just compare the pointer of this term with the argument
+  // t. This is because we actually need to use this compare method for equality
```
yoni206, line 63:
```suggestion
  // This is because we actually need to use this compare method for equality
```

src/logging_term.cpp
```diff
@@ -46,6 +46,25 @@ LoggingTerm::~LoggingTerm() {}
...
+  // when looking up this term in the hash table to perform hash-consing
```
yoni206, line 64:
```suggestion
  // when looking up this term in the hash table to perform hash-consing.
```

src/logging_term.cpp
```diff
@@ -46,6 +46,25 @@ LoggingTerm::~LoggingTerm() {}
...
+  // Thus, we cannot count on the pointers being equal for this term yet
+  // only for the children which have already been resolved in the hash table.
```
yoni206, line 66: Can u rephrase a bit?

src/logging_term.cpp
yoni206, line 64: can you clarify that this is done when constructing the LoggingTerm?

src/logging_term.cpp
makaimann: Yep, hopefully made it clearer with 2dab303f5364d11b408b9b55268d7f7d5415d03e

src/logging_term.cpp
makaimann: Yep, thanks! Done with 2dab303f5364d11b408b9b55268d7f7d5415d03e

src/logging_term.cpp
makaimann: Fixed in 2dab303f5364d11b408b9b55268d7f7d5415d03e

PR #84, opened 2020-07-10
Fix for constant array term iteration in CVC4 + tests

makaimann:
Smt-switch uses term iteration to get the constant array base of a constant array. For example, if `t` is a constant array, the base is `*(t->begin())`. CVC4 does not have this same representation. Previously, term iteration for the CVC4 backend used the `::CVC4::api::Term::const_iterator` directly. This PR would instead pass the parent term and a position and use `operator[]` of the term object to get the children. There's a special case for constant arrays where the end position is increased by one and the first iterator returns the base array.

### DIFF
--- a/cvc4/include/cvc4_term.h
+++ b/cvc4/include/cvc4_term.h
@@ -45,7 +45,7 @@ namespace smt {
   {
   public:
    CVC4TermIter(::CVC4::api::Term term, uint32_t p = 0) : term(term), pos(p){};
-   CVC4TermIter(const CVC4TermIter & it) { term = term; };
+   CVC4TermIter(const CVC4TermIter & it) { term = it.term; };
    ~CVC4TermIter(){};
    CVC4TermIter & operator=(const CVC4TermIter & it);
    void operator++() override;
@@ -92,4 +92,3 @@ namespace smt {
 
 
 }
-
--- a/cvc4/include/cvc4_term.h
+++ b/cvc4/include/cvc4_term.h
@@ -44,22 +44,22 @@ namespace smt {
   class CVC4TermIter : public TermIterBase
   {
   public:
-    CVC4TermIter(const ::CVC4::api::Term::const_iterator term_it)
-      : term_it(term_it){};
-    CVC4TermIter(const CVC4TermIter & it) { term_it = it.term_it; };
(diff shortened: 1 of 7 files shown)


cvc4/include/cvc4_term.h
```diff
@@ -44,22 +44,22 @@ namespace smt {
...
-    CVC4TermIter(const CVC4TermIter & it) { term_it = it.term_it; };
-    ~CVC4TermIter() {};
-    CVC4TermIter & operator=(const CVC4TermIter & it);
-    void operator++() override;
-    const Term operator*() override;
-    TermIterBase * clone() const override;
-    bool operator==(const CVC4TermIter & it);
-    bool operator!=(const CVC4TermIter & it);
+   CVC4TermIter(::CVC4::api::Term term, uint32_t p = 0) : term(term), pos(p){};
+   CVC4TermIter(const CVC4TermIter & it) { term = term; };
```
aniemetz, line 48:
```suggestion
   CVC4TermIter(const CVC4TermIter & it) { term = it.term; };
```

cvc4/src/cvc4_term.cpp
```diff
@@ -119,48 +119,57 @@ CVC4::api::TermHashFunction termhash;
...
 bool CVC4TermIter::operator==(const CVC4TermIter & it)
 {
-  return term_it == it.term_it;
+  return term == it.term && pos == it.pos;
 }
 
 bool CVC4TermIter::operator!=(const CVC4TermIter & it)
 {
-  return term_it != term_it;
+  return term != term || pos != it.pos;
```
aniemetz, line 166:
```suggestion
  return term != it.term || pos != it.pos;
```

cvc4/include/cvc4_term.h
makaimann: oops, good catch. That's weird that it was passing tests, but I guess it's cause the term was always the same for normal iteration.

makaimann:
Okay, I added a dedicated test for the issue you found. I'll add more generic TermIter tests in a separate PR.

aniemetz (approved):

makaimann:
Thanks for the review!

PR #91, opened 2020-08-05
Hash string representation of mathsat simple type

makaimann:
MathSAT does not have ids for types. Thus the hash implementation for sorts in the MathSAT backend is custom and performed at the smt-switch level. It first determines the general type (e.g. Int, Array, etc...) and then incorporates the sort parameters into the hash.

However, MathSAT doesn't currently have a way to check if a given type is a simple type (declared sort). Thus, to implement hash for simple types, we just hash the string representation of the type.

Note: one danger here is that when new types are added, this could make it so the developer forgets to add a specialized hash for that type, and instead it always relies on the string. This should always still work but could possibly have a slight performance impact.


makaimann:
All right, I added a few tests and used the `msat_type` pointer for the MathSAT sort hash and it seems to be working. What do you think? Should this be safe?

PR #92, opened 2020-08-05
changes to unsat_core method

ahmed-irfan:
This PR makes the following update
- instead of returning a TermVec, the get_unsat_core method takes a reference of either a TermVec or a UnorderedTermSet as an argument. 

This change will make the method more memory-friendly.

### DIFF
--- a/btor/include/boolector_solver.h
+++ b/btor/include/boolector_solver.h
@@ -62,7 +62,6 @@ class BoolectorSolver : public AbsSmtSolver
   Term get_value(const Term & t) const override;
   UnorderedTermMap get_array_values(const Term & arr,
                                     Term & out_const_base) const override;
-  void get_unsat_core(TermVec & out) override;
   void get_unsat_core(UnorderedTermSet & out) override;
   Sort make_sort(const std::string name, uint64_t arity) const override;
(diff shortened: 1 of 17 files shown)


makaimann:
Looks like the Travis errors are just because of this: https://github.com/makaimann/smt-switch/blob/90471ac838667745da9a4c3bdaf2465dc7551b88/tests/cvc4/cvc4-printing.cpp#L92

ahmed-irfan:
> Just FYI -- looks like the Travis errors are just because of this:
> 
> https://github.com/makaimann/smt-switch/blob/90471ac838667745da9a4c3bdaf2465dc7551b88/tests/cvc4/cvc4-printing.cpp#L92
> 
> Also, it appears there is a test for the unsat core in python which passed. I'll give a closer review after the vector version has been removed (as discussed offline).
> 
> Thanks for implementing this!

FIxed

makaimann:
I'm going to merge this now. Just a reminder: there will need to be changes in Pono, but also in https://github.com/yoni206/lazybv2int

ahmed-irfan:
Thanks for the reminder. I can take care of those changes

PR #93, opened 2020-08-06
Have optional user-provided cache argument to IdentityWalker

makaimann:


ahmed-irfan:
How about: instead of exposing the internal cache, identity walker can work also on the user provided cache?

makaimann:
> How about: instead of exposing the internal cache, identity walker can work also on the user provided cache?

Yeah that could make sense: so there would be two caches? One internal and one provided by the user? Actually, we could have the restriction that the user-provided cache be fully populated before passing it. Then we can just copy values from it to the internal cache. What do you think?

ahmed-irfan:
I am thinking to have both caches: internal and user-provided. The user-provided is a pointer initialized to NULL. So, during the traversal, if it is not NULL then we use the user-provided cache. Otherwise, we use the internal one.

makaimann:
Makes sense! Just made those changes. Would you mind reviewing when you get a chance?

include/identity_walker.h
```diff
@@ -44,36 +44,66 @@ enum WalkerStepResult
...
+                smt::UnorderedTermMap * ext_cache = nullptr)
+     : solver_(solver), clear_cache_(clear_cache), ext_cache_(ext_cache){};
 
-  /** Visit a term and all its subterms in a pre- and post-order traversal
-   *  @param term the term to visit
-   *  @return the term after visiting (returns the value of cache[term]
-   *     -- if it has been cached and returns term otherwise)
-   */
-  smt::Term visit(smt::Term & term);
+ /** Visit a term and all its subterms in a pre- and post-order traversal
```
ahmed-irfan, line 61: pre- and post-order ? Isn't it one?

include/identity_walker.h
```diff
@@ -44,36 +44,66 @@ enum WalkerStepResult
...
+  *  @return true iff there is a cache hit
+  */
+ bool query_cache(const Term & key, Term & out) const;
+
+ /** Populate the cache. Automatically uses ext_cache_ if non-null.
+  *  It will overwrite the existing mapping if the key is already in the cache
+  *  @param key the key term
+  *  @param val the value term
+  */
+ void populate_cache(const Term & key, const Term & val);
```
ahmed-irfan, line 95: can we change the name of this method? It sounds confusing to me. How about `save_in_cache`?

makaimann: Yep! Changed with e1fc7b40e57e7d5ac26c9ed053854ea21fc6a5c1

include/identity_walker.h
makaimann: Updated the comment with b1ff8dd5035c096443dab83b266eaf12c25d925a. Is that clearer?

ahmed-irfan: yup :)

PR #96, opened 2020-08-24
constant rationals support in transfer_term

yoni206:
When transferring a term, constant rationals are created using a string of the form `(/ a b)`, while some solvers expect a string of the form `a / b`.
Specifically, `mathsat` expects `a / b` while `cvc4` can handle both forms. `yices2` currently does not transfer terms (does not have a `begin` function for the iterator) and `boolector` does not support arith. Hence I just used the infix notation.

### DIFF
--- a/src/term_translator.cpp
+++ b/src/term_translator.cpp
@@ -279,11 +279,10 @@ std::string TermTranslator::mathsatize_rational(const std::string smtlib) const 
     int ind_of_up_end = smtlib.find_first_of(' ', ind_of_up_start);
   assert(ind_of_up_end != std::string::npos);
   ind_of_up_end -= 1;
-  int ind_of_down_start = ind_of_up_end + 2; 
+  int ind_of_down_start = ind_of_up_end + 2;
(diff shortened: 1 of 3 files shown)


src/term_translator.cpp
```diff
@@ -265,6 +266,29 @@ Term TermTranslator::transfer_term(const Term & term, const SortKind sk)
...
+  // smtlib: (/ up down)
+  // ind -- index
+  std::string op;
+  int ind_of_up_start = smtlib.find_first_of("/");
+  if (ind_of_up_start == std::string::npos) {
+    return smtlib;
+  }
+  ind_of_up_start += 2;
+  op = "/";
+    int ind_of_up_end = smtlib.find_first_of(' ', ind_of_up_start);
```
ahmed-irfan, line 279: format

yoni206:
thanks. fixed in https://github.com/makaimann/smt-switch/pull/96/commits/546932a71e834216a8bd5c10aec29756c1ca2b96

PR #97, opened 2020-08-28
Add get_arity and notion of a sort constructor

makaimann:
Adds a `get_arity` method to sorts for obtaining the arity of an uninterpreted sort. It also adds a new `SortKind` called `UNINTERPRETED_CONS` for an uninterpreted sort constructor. This is to distinguish between a sort declared with non-zero arity before (`UNINTERPRETED_CONS`) and after (`UNINTERPRETED`) it has been applied to subsort arguments.


makaimann:
It appears that `getUninterpretedSortParamSorts` is broken in CVC4. Needs more investigation.

PR #102, opened 2020-09-05
Option to include BVAnd/BVOr ops in conjunctive/disjunctive partition

makaimann:
Allows include `bvand` or `bvor` in partition util functions if they are over one-bit bit-vectors. This can be used with boolector which represents booleans as bit-vectors of size one.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -55,10 +55,24 @@ namespace smt {
 // term helper methods
 void op_partition(smt::PrimOp o, const smt::Term & term, smt::TermVec & out);
 
+/** Populates a vector with conjuncts from a term
+ *  @param the term to partition
+ *  @param out the output vector
+ *  @param include_bvand also split on BVAnd over 1-bit signals
+ *         Note: this is mainly for use with Boolector which treats
+ *         booleans as 1-bit bit-vectors. Using this option on a term
+ *         that is known to be boolean will give you the expected
+ *         partition.
+ */
 void conjunctive_partition(const smt::Term & term,
                            smt::TermVec & out,
                            bool include_bvand = false);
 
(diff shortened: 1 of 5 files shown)


include/utils.h
```diff
@@ -50,10 +50,19 @@ inline void Log(std::string msg)
...
 
+namespace smt {
+
 // term helper methods
-void conjunctive_partition(const smt::Term &term, smt::TermVec &out);
+void op_partition(smt::PrimOp o, const smt::Term & term, smt::TermVec & out);
+
+void conjunctive_partition(const smt::Term & term,
+                           smt::TermVec & out,
+                           bool include_bvand = false);
```
ahmed-irfan, line 60: perhaps, it is good to add comment here about `include_bvand`. Specially, it is important to mention about 1-bit ...

include/utils.h
```diff
@@ -50,10 +50,19 @@ inline void Log(std::string msg)
...
 
-void disjunctive_partition(const smt::Term &term, smt::TermVec &out);
+void disjunctive_partition(const smt::Term & term,
+                           smt::TermVec & out,
+                           bool include_bvor = false);
```
ahmed-irfan, line 64: ditto

ahmed-irfan (changes requested):
Overall LGTM. Only requested minor additions about comments

include/utils.h
makaimann: Good point!

makaimann:
Thanks for the reviews! Addressed in b05fe3759cf666cc56c79a3b998d808f2568da43

ahmed-irfan (approved):

PR #106, opened 2020-10-05
Use `-undefined dynamic_lookup` when `Py_ENABLE_SHARED` is 0

leonardt:
Fixes https://github.com/upscale-project/pono/issues/92

It seems that the `target_link_libraries_with_dynamic_lookup` logic
doesn't properly discover this requirement on my version of macOS
(10.15.6) and conda (conda 4.8.3).

This uses the cmake command from
https://github.com/conda-forge/python-feedstock/issues/272 to read
Py_ENABLE_SHARED.  If this config var isn't set, we use `-undefined
dynamic_lookup` rather than linking the python library directly

### DIFF
--- a/cmake/FindPythonLibs.cmake
+++ b/cmake/FindPythonLibs.cmake
@@ -268,6 +268,7 @@ function(python_extension_module _target)
     # linux doesn't need to link against python libraries
     if (APPLE)
         # determine linkage of python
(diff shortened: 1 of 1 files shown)


leonardt:
Closing in favor of https://github.com/makaimann/smt-switch/pull/107

PR #107, opened 2020-10-05
Rely on scikit-build CMake files for Cython bindings

makaimann:
Updates the instructions and CMake to use the scikit-build version of CMake files for Cython bindings from: https://github.com/scikit-build/scikit-build/commit/732f4c3f2a7e08c9154df4e5f7ddb9cd01d5044a. The setup script pins a particular commit, but can easily be updated. Like other dependencies, the idea is to not have the build fail because of upstream changes, but also allow easy updates to keep in line.


makaimann:
Note: Updated this PR to pull https://github.com/scikit-build/scikit-build for the CMake files for Python/Cython extensions, instead of just copying the files.

PR #109, opened 2020-10-09
CVC4 interpolation

yoni206:
This PR adds support for cvc4's interpolation api.


tests/CMakeLists.txt
```diff
@@ -123,10 +123,10 @@ target_link_libraries(test-int gtest_main)
 target_link_libraries(test-int test-deps)
 add_test(NAME test-int_test COMMAND test-int)
 
-add_executable(test-dt "${PROJECT_SOURCE_DIR}/tests/test-dt.cpp")
-target_link_libraries(test-dt gtest_main)
-target_link_libraries(test-dt test-deps)
-add_test(NAME test-dt_test COMMAND test-dt)
+#add_executable(test-dt "${PROJECT_SOURCE_DIR}/tests/test-dt.cpp")
```
makaimann, line 126: Can this test be enabled again?

makaimann (changes requested):
It's looking pretty good to me! Just left a minor comment. Also, should we run CVC4 in the generic interpolation tests?

The tests are here: https://github.com/makaimann/smt-switch/blob/master/tests/test_itp.cpp

I think the only places you need to make changes are adding an interpolator enum for CVC4 here: https://github.com/makaimann/smt-switch/blob/c64238272da30701e68db9cc8a04fee068a805e2/include/solver_enums.h#L36

Then you tell this function how to instantiate a CVC4 interpolator here: https://github.com/makaimann/smt-switch/blob/c64238272da30701e68db9cc8a04fee068a805e2/tests/available_solvers.cpp#L125

And finally, list the `CVC4_INTERPOLATLOR` enum here:
https://github.com/makaimann/smt-switch/blob/c64238272da30701e68db9cc8a04fee068a805e2/tests/available_solvers.cpp#L142

yoni206:
Thanks for the feedback! I believe all is addressed. Please let me know if you have any other comments.

PR #110, opened 2020-10-12
Parameterize tests by configurations, not enums

yoni206:
This PR introduces a new `struct`, `SolverConfiguration` for the testing infra-structure.
A `SolverConfiguration` consists of a `SolverEnum` and a Boolean flag that states whether this solver is a `LoggingSolver`.
Tests are now parameterized by `SolverConfiguration`.
The `_LOGGING` enums are deleted.

### DIFF
--- a/tests/available_solvers.cpp
+++ b/tests/available_solvers.cpp
@@ -160,8 +160,6 @@ std::vector<SolverConfiguration> available_interpolator_configurations() {
     SolverConfiguration sc;
     sc.solver_enum = e;
     sc.is_logging_solver = false;
-    //TODO do we ignore logging_interpolator? 
-    // seems like this is the case currently on master?
     result.push_back(sc);
   }
   return result;
--- a/tests/available_solvers.cpp
+++ b/tests/available_solvers.cpp
@@ -124,8 +124,8 @@ SmtSolver create_solver(SolverEnum se)
 
 SmtSolver create_solver(SolverConfiguration sc)
 {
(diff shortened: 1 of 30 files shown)


tests/available_solvers.cpp
```diff
@@ -149,38 +125,20 @@ SmtSolver create_interpolating_solver(SolverEnum se)
...
   {
-    const std::unordered_set<SolverAttribute> & se_attrs =
-        get_solver_attributes(se);
-
-    if (se_attrs.find(LOGGING) != se_attrs.end())
-    {
-      enums.push_back(se);
-    }
+    SolverConfiguration sc;
+    sc.solver_enum = e;
```
makaimann, line 135: I don't feel too strongly about this, but it does seem a bit strange to assign to the member variables this way and implicitly rely on the copy constructor when adding to the vector (so we're not just putting the same object in `configs` twice). An alternative is to add a constructor to the struct that lets you create the object with initialized values: `SolverConfiguration(enum, logging)`. Might be a bit cleaner, even though it's not really different semantically. What do you think?

makaimann (approved):
Looks really good to me. Thanks for working on this! Left a couple minor comments, but otherwise I think it's good to go once the tests pass. We talked elsewhere about the python tests -- let me know if I can help with those.

makaimann:
Looks good, thanks for the updates and for this refactor! I think it's much cleaner now.

PR #120, opened 2020-11-05
Take a const reference in Sort compare

makaimann:
I noticed recently that the `Sort::compare` method takes a `const Sort`, when it could take a `const Sort &`. The reference is better because it avoids incrementing/decrementing the reference counter of the `Sort` (which is a `shared_ptr`). It's a minor change but had to be updated for all backends.

Note, this will need to be adjusted for other backends that haven't been merged yet @yoni206 and @lstuntz.


amaleewilson (commented):
Looks good to me.

amaleewilson (approved):

makaimann:
Thanks for the review!

PR #122, opened 2020-11-06
Unsatcore reducer

ahmed-irfan:


src/utils.cpp
```diff
@@ -126,4 +129,120 @@ void get_free_symbolic_consts(const smt::Term &term, smt::TermVec &out)
...
+                                               TermVec &out_red,
+                                               TermVec *out_rem,
+                                               unsigned iter,
+                                               unsigned rand_seed)
+{
+  TermVec bool_assump, tmp_assump;
+  UnorderedTermMap to_ext_assump;
+  TermVec cand_res;
+  for (auto a : assump) {
+    Term t = to_reducer_.transfer_term(a);
```
makaimann, line 156: One thing to mention here is that the `TermTranslator` implicitly assumes that it is only translating from one solver. So, this `UnsatCoreReducer` is tied to a particular `ext_solver`. That's fine as long as that's the behavior you expect. Otherwise, you could clear the `TermTranslator` cache each time this method is called, but then you might have a performance hit.

makaimann (approved):
LGTM, thanks for adding this! I only left one comment for you to consider.

src/utils.cpp
ahmed-irfan: that's a good point. To address this, I created a method to reset the cache of the translator.

makaimann:
Thanks for the updates! Looks like there's a few conflicts after merging #121. Would you mind resolving those when you get a chance?

PR #132, opened 2020-11-21
refator get_free_symbols

ahmed-irfan:


src/utils.cpp
```diff
@@ -105,8 +105,9 @@ void disjunctive_partition(const smt::Term & term,
   }
 }
 
-void get_free_symbolic_consts(const smt::Term & term,
-                              smt::UnorderedTermSet & out)
+void get_kind_terms(const smt::Term & term,
```
makaimann, line 108: Maybe there's a better name, like `get_matching_terms`? I think it might be nice to expose this to the API actually and document it in the header file. Then the `get_free_*` are just convenient special cases.

makaimann (approved):
Very cool! I like this refactor a lot. I mentioned this in a comment, already but I think it would be cool to expose this for the public API. I could see it being very useful for people. What do you think?

ahmed-irfan:
Thanks for the review! 
Made the suggested changes.

PR #139, opened 2020-12-14
Add get_sequence_interpolants function for interpolating solvers

makaimann:


ahmed-irfan (approved):
LGTM

makaimann:
Thanks for the review @ahmed-irfan ! Sorry for not adding it earlier, but I made one minor improvement. The latest commit just makes it so that there can be interpolation failures for some of the sequence interpolants and it will still finish. It will just use null terms in the vector (so that you can still figure out which part of the sequence the non-null interpolants line up with) and return unknown instead of unsat.

PR #143, opened 2020-12-18
Move constant strings for printing solver to utils

yoni206:
The constant strings for SMT-LIB commands will be shared between the printing solver and the generic solver.
As a first step, this PR moves them from the printing solver to the utils file.


makaimann (commented):
What do you think about creating a new header for this? It could just be another generic name like `constant_utils.h` or something like that? I'm not sure it complete belongs in `utils.h`. At least based on how we've been using it so far. Basically, `utils` has utilities for a user and I'm not sure these need to be so public facing.

yoni206:
Yup makes sense. Pushed. How does it look now?

makaimann (approved):
LGTM, thanks for the changes!

PR #145, opened 2020-12-21
Changing SMT-LIB constants to strings instead of macros

yoni206:
Actually, the planned generic solver uses these macros as strings, not chars. So this PR makes them to be string constants instead of macros.


makaimann (approved):
LGTM! I'll merge it after Travis finishes.

PR #146, opened 2020-12-23
Header for Generic Sorts

yoni206:
This PR introduces a header file (no implementations) for generic sorts, that will be used be generic solvers.
It follows a similar structure to that of a logging sort.

I've also included an empty test file that includes `generic_sort.h`, as otherwise the header is not compiled and errors are not detected. The test file will be filled gradually as implementations are added.

~~The `cmake` file points to the header file. This will be changed once implementations are added.~~
Nothing is added to the `cmake` files, as the header file is compiled thanks to its inclusion in the test.

### DIFF
--- a/include/generic_sort.h
+++ b/include/generic_sort.h
@@ -56,7 +56,7 @@ class GenericSort : public AbsSort
         "get_width not implemented by GenericSort");
   }
 
-  uint64_t get_arity() const override
+  size_t get_arity() const override
   {
     throw NotImplementedException(
         "get_width not implemented by GenericSort");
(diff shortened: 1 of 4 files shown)


include/generic_sort.h
```diff
@@ -0,0 +1,190 @@
...
+   *   implemented by sub-classes 
+   */
+
+  uint64_t get_width() const override
+  {
+    throw NotImplementedException(
+        "get_width not implemented by GenericSort");
+  }
+
+  uint64_t get_arity() const override
```
makaimann, line 59: This should be `size_t` actually -- that should also fix the Travis build. It is strange that `get_width` returns `uint64_t` and `get_arity` returns `size_t`. You're welcome to fix that but it's not your responsibility, I can go through and pick one later.

makaimann (approved):
LGTM, thanks! Just made one comment which should also fix the Travis build.

PR #147, opened 2020-12-27
Generic Sort -- Some Implementations

yoni206:
This PR adds implementations for about 50% of the generic sort interface.
It also extends the test file accordingly.

### DIFF
--- a/src/generic_sort.cpp
+++ b/src/generic_sort.cpp
@@ -32,6 +32,7 @@ size_t GenericSort::hash() const {
 
 string GenericSort::to_string() const {
   return compute_string();
+}
 
 string GenericSort::compute_string() const {
     if (get_sort_kind() == SortKind::ARRAY) {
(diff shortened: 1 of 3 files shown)


src/generic_sort.cpp
```diff
@@ -0,0 +1,161 @@
...
+             && (get_elemsort() == s->get_elemsort());
+    }
+    case FUNCTION:
+    {
+      SortVec domain_sorts = get_domain_sorts();
+      SortVec other_domain_sorts = s->get_domain_sorts();
+      Sort return_sort = get_codomain_sort();
+      Sort other_return_sort = s->get_codomain_sort();
+
+      if (domain_sorts.size() != other_domain_sorts.size()
```
makaimann, line 111: I only learned this recently, but `vector` actually has `operator==` so you can just compare `domain_sorts == other_domain_sorts` if you'd like. You could compare the return sorts first since it's the cheaper operation also. I assume that the `vector` `operator==` is smart enough to check that the sizes match first and whatnot, but I didn't verify that.

makaimann (approved):
Looks good to me, thanks for working on this! Just left a minor comment which you don't need to address.

PR #149, opened 2020-12-29
Add method to get unique Term ID

makaimann:
All solvers that I'm aware of support getting a unique id for a term. For many solvers, there is no distinction between the hash and the unique id, but in smt-switch we will have both to support solvers where there is a difference.

### DIFF
--- a/tests/unit/unit-term-id.cpp
+++ b/tests/unit/unit-term-id.cpp
@@ -1,0 +1,86 @@
+
+/*********************                                           */
+/*! \file unit-term-id.cpp
+** \verbatim
+** Top contributors (to current version):
+**   Makai Mann
+** This file is part of the smt-switch project.
+** Copyright (c) 2020 by the authors listed in the file AUTHORS
+** in the top-level source directory) and their institutional affiliations.
+** All rights reserved.  See the file LICENSE in the top-level source
+** directory for licensing information.\endverbatim
+**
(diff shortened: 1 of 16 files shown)


makaimann:
FYI @lstuntz and @yoni206. Hopefully this PR won't complicate your respective backends too much. If it becomes a problem, you can just throw an exception for now and exclude it from the test.

tests/unit/unit-term-id.cpp
```diff
@@ -0,0 +1,86 @@
...
+    s->set_opt("produce-models", "true");
+    s->set_opt("incremental", "true");
+
+    boolsort = s->make_sort(BOOL);
+    bvsort = s->make_sort(BV, 4);
+    a = s->make_symbol("a", boolsort);
+    b = s->make_symbol("b", boolsort);
+    x = s->make_symbol("x", bvsort);
+    y = s->make_symbol("y", bvsort);
+    one = s->make_term(1, bvsort);
```
ahmed-irfan, line 47: can we make another term with the same constant value. 

`second_one = s->make_term(1, bvsort);`

and check its equivalence with `one`.

Does that make sense?

makaimann: Yeah that's a great idea, thanks! Added with dc744e79c17f011b3877225118d8f177af359d05

ahmed-irfan (approved):
LGTM

PR #150, opened 2020-12-30
get_ops oracle

yoni206:
This PR introduces a `get_ops` function to `utils`, that computes the set of operators in a given term.
The PR also includes some tests for this new function, as well as for `get_free_symbolic_constants`.
Since the computed set of operators is stored in an `unordered_set`, I also added a trivial `hash` function that relies on the string representation of the operators.

### DIFF
--- a/include/ops.h
+++ b/include/ops.h
@@ -156,7 +156,7 @@ std::ostream& operator<<(std::ostream& output, const Op o);
 // defining hash for old compilers
 namespace std
 {
-  // specialize the hash template
+  // specialize the hash template for PrimOp
   template<>
     struct hash<smt::PrimOp>
     {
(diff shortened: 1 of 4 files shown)


makaimann (approved):
LGTM, thanks!

include/ops.h
```diff
@@ -162,5 +165,17 @@ namespace std
...
+
+  // specialize the hash template for Op
+  template<>
+    struct hash<smt::Op>
+    {
+      size_t operator()(const smt::Op o) const
+      {
+        // The hash function for op computes the string hash
+        std::hash<std::string> str_hash;
+        return str_hash(o.to_string());
```
makaimann, line 177: One comment: if we ever decide this hash is too slow (which I doubt), then we could always just default to using the hash of the `PrimOp`. That actually seems like a reasonable thing to do to me, because there's only a few indexed operators, and I don't think it's so bad if those are all kept in the same table entry, e.g. you'd have:
`hash(Extract) -> [(_ extract 0 0), (_ extract 8 4), ...]`

Not saying we need to make that change, just mentioning it.

yoni206: Yes I agree that this can be optimized, but also doubt it will become a bottleneck (it is always O(1), except for maybe crazy uninterpreted sorts with arities > 1 built on top of each other).

Issue #151, opened 2020-12-31
Push Pop unbalance in `reduce_assump_unsatcore`


zhanghongce:
https://github.com/makaimann/smt-switch/blob/53ad7bd0a88c6e8cb6e8989540fc7899a668b6f2/src/utils.cpp#L209

Hi I'm wondering, if you directly return after `sat`, wouldn't you cause unbalanced push/pop issue?
Thanks!

zhanghongce:
Address by

PR #152, opened 2020-12-31
Adding an unsat core reduce function

zhanghongce:
This is to iterate through the unsat assumptions. This is used in SyGuS PDR and it guarantees to be minimal (but not minimum)

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -121,6 +121,24 @@ public:
                                smt::TermVec *out_rem = NULL,
                                unsigned iter = 0,
                                unsigned rand_seed = 0);
+  
+
+  /** The additional method to reduce the assump (vector of assumptions). The method
+   *  assumes that the conjunction of the formula and assump is unsatisfiable.
+   *  This will iterate through the assump and requires at most size(assump) query
+   *  @param input formula
+   *  @param input vector of assumptions
+   *  @param output vector for the reduced assumptions
+   *  @param output vector for the removed assumptions
+   *  @param iter is the number of iterations done in the method. Default is 0,
+   *         and it means that the result in out_red will be minimal.
+   */
+  bool linear_reduce_assump_unsatcore(
+                               const smt::Term &formula,
+                               const smt::TermVec &assump,
+                               smt::TermVec &out_red,
+                               smt::TermVec *out_rem = NULL,
+                               unsigned iter = 0);
 
   /** This clears the term translation cache. Note, term translator is used to
    *  translate the terms of the external solver to the
--- a/include/utils.h
+++ b/include/utils.h
@@ -126,12 +126,16 @@ public:
   /** The additional method to reduce the assump (vector of assumptions). The method
    *  assumes that the conjunction of the formula and assump is unsatisfiable.
    *  This will iterate through the assump and requires at most size(assump) query
+   *  Note: this function assumes that there are no duplicate assumptions from the
+   *  second input vector.
    *  @param input formula
    *  @param input vector of assumptions
    *  @param output vector for the reduced assumptions
    *  @param output vector for the removed assumptions
    *  @param iter is the number of iterations done in the method. Default is 0,
    *         and it means that the result in out_red will be minimal.
+   *  returns false if the formula conjoined with the assump is satisfiable,
+   *          otherwise returns true
    */
   bool linear_reduce_assump_unsatcore(
                                const smt::Term &formula,
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -272,10 +272,12 @@ bool UnsatCoreReducer::linear_reduce_assump_unsatcore(
     return true;
   }
 
+  UnorderedTermMap label_to_cand_;
   for (const auto & a : cand_res) {
     Term l = label(a);
     reducer_->assert_formula(reducer_->make_term(Implies, l, a));
     bool_assump.push_back(l);
+    label_to_cand_.emplace(l, a);
   }
 
   unsigned cur_iter = 0;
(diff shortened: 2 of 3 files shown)


zhanghongce:
I'm still working on this. Especially, there are questions about the expected behavior of duplicate assumptions.

makaimann (approved):
LGTM, thanks for the PR!

makaimann:
Marking `do not merge` until you (@zhanghongce) say you're done working on it. If you're worried about duplicate assumptions, you could consider de-duplicating with an `UnorderedTermSet` first.

zhanghongce:
Thanks! I think it is now ready for review.

src/utils.cpp
```diff
@@ -246,6 +246,104 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+    return false;
+  }
+  assert(r.is_unsat());
+  TermVec bool_assump_query;
+  UnorderedTermSet core_set(bool_assump.begin(), bool_assump.end());
+
+  while (cur_iter <= iter && assump_pos < bool_assump.size()) {
+    cur_iter = iter > 0 ? cur_iter+1 : cur_iter;
+    bool_assump_query = bool_assump;
+    // std::cout << "try erasing : #" << assump_pos << std::endl;
```
makaimann, line 295: minor: can delete this line

src/utils.cpp
```diff
@@ -246,6 +246,104 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+        auto bool_assump_pos = bool_assump.begin();
+        while(bool_assump_pos != bool_assump.end()) {
+          // if not in the unsat core, remove it
+          if ( (bool_assump_pos - bool_assump.begin() == assump_pos) ||
+              core_set.find(*bool_assump_pos) == core_set.end() )
+            bool_assump_pos = bool_assump.erase(bool_assump_pos);
+          else
+            ++ bool_assump_pos;  
+        }
+        // std::cout << "unsat, bool_assump size : " << bool_assump.size() << std::endl;
```
makaimann, line 319: minor: can remove this line

src/utils.cpp
```diff
@@ -246,6 +246,104 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+      }
+      assert(bool_assump.size() <= bool_assump_query.size());
+      // we don't need to change assump_pos, because the size of
+      // bool_assump is reduced by at least one, so in the next round
+      // it is another element sitting at this location
+    } // end of else if (unsat)
+    // at this point, bool_assump should be the output
+  }
+
+  // std::cout << "iteration complete  :  core_set size = " << core_set.size() << std::endl;
```
makaimann, line 329: minor: can remove this line

src/utils.cpp
```diff
@@ -246,6 +246,104 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+  unsigned cur_iter = 0;
+  unsigned assump_pos = 0;
+  r = reducer_->check_sat_assuming(bool_assump);
+  if (r.is_sat()) {
+    reducer_->pop();
+    return false;
+  }
+  assert(r.is_unsat());
+  TermVec bool_assump_query;
+  UnorderedTermSet core_set(bool_assump.begin(), bool_assump.end());
```
makaimann, line 290: do you actually need to populate the set here before the loop?

tests/test-unsat-core-reducer.cpp
```diff
@@ -60,6 +60,23 @@ TEST_P(UnsatCoreReducerTests, UnsatCoreReducer)
...
+  UnsatCoreReducer uscr(r);
+
+  Term a = s->make_symbol("a", boolsort);
+  Term b = s->make_symbol("b", boolsort);
+  Term formula = s->make_term(And, a, b);
+  TermVec assump({ s->make_term(Not, a), s->make_term(Not, b) });
+  TermVec red, rem;
+
+  uscr.linear_reduce_assump_unsatcore(formula, assump, red, &rem);
+  EXPECT_TRUE(red.size() == 1);
```
makaimann, line 74: This isn't too important, but it's more idiomatic to use:
`EXPECT_EQ(<arg1>, <arg2>)` and
`EXPECT_NE(<arg1>, <arg2>)`

makaimann (approved):
Looks good, thanks @zhanghongce! I just left a few minor comments that you can address or not.

src/utils.cpp
zhanghongce: This was for the case that not a single assumption can be removed. (In that case no UNSAT core extraction is executed)
But maybe I can write in another way that won't rely on the final `core_set`... let me change to that ...

zhanghongce:
The commit 013f199 will resolve #151

makaimann:
> The commit [013f199](https://github.com/makaimann/smt-switch/commit/013f199fde65fb8a756af4ec1a31b731284db57e) will resolve #151

Good catch! That's definitely a problem.

makaimann:
@ahmed-irfan do you want to make a pass when you have a chance? Looks good to me, but a second pair of eyes is always good.

include/utils.h
```diff
@@ -121,6 +121,24 @@ class UnsatCoreReducer {
...
+  /** The additional method to reduce the assump (vector of assumptions). The method
+   *  assumes that the conjunction of the formula and assump is unsatisfiable.
+   *  This will iterate through the assump and requires at most size(assump) query
+   *  @param input formula
+   *  @param input vector of assumptions
+   *  @param output vector for the reduced assumptions
+   *  @param output vector for the removed assumptions
+   *  @param iter is the number of iterations done in the method. Default is 0,
+   *         and it means that the result in out_red will be minimal.
+   */
```
ahmed-irfan, line 135: add comment about the result

src/utils.cpp
```diff
@@ -246,6 +248,107 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+      // we can remove this assumption
+      assert(r.is_unsat());
+
+      UnorderedTermSet core_set;
+      reducer_->get_unsat_core(core_set);
+      { // remove from bool_assump
+        auto bool_assump_pos = bool_assump.begin();
+        while(bool_assump_pos != bool_assump.end()) {
+          // if not in the unsat core, remove it
+          if ( core_set.find(*bool_assump_pos) == core_set.end() )
```
ahmed-irfan, line 313: just a style thing: add brackets here

src/utils.cpp
```diff
@@ -246,6 +248,107 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+            bool_assump_pos = bool_assump.erase(bool_assump_pos);
+          else
```
ahmed-irfan, line 315: ditto

src/utils.cpp
```diff
@@ -246,6 +248,107 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+      ++ assump_pos;
+    } else {
+      // we can remove this assumption
+      assert(r.is_unsat());
+
+      UnorderedTermSet core_set;
+      reducer_->get_unsat_core(core_set);
+      { // remove from bool_assump
+        auto bool_assump_pos = bool_assump.begin();
+        while(bool_assump_pos != bool_assump.end()) {
```
ahmed-irfan, line 311:
```suggestion
        while (bool_assump_pos != bool_assump.end()) {
```

src/utils.cpp
```diff
@@ -246,6 +248,107 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+  while (cur_iter <= iter && assump_pos < bool_assump.size()) {
+    cur_iter = iter > 0 ? cur_iter+1 : cur_iter;
+    bool_assump_query = bool_assump;
+    bool_assump_query.erase(bool_assump_query.begin()+assump_pos);
+    r = reducer_->check_sat_assuming(bool_assump_query);
+    if (r.is_sat()) {
+      // we cannot remove this assumption, then try next one
+      ++ assump_pos;
+    } else {
+      // we can remove this assumption
```
ahmed-irfan, line 304: where do we remove this assumption? `assump_pos`

src/utils.cpp
```diff
@@ -246,6 +248,107 @@ bool UnsatCoreReducer::reduce_assump_unsatcore(const Term &formula,
...
+bool UnsatCoreReducer::linear_reduce_assump_unsatcore(
+                              const smt::Term &formula,
+                              const smt::TermVec &assump,
+                              smt::TermVec &out_red,
+                              smt::TermVec *out_rem,
+                              unsigned iter)
+{
+  TermVec bool_assump, tmp_assump;
+  UnorderedTermMap to_ext_assump;
+  TermVec cand_res;
```
ahmed-irfan, line 260: maybe rename this to local_assump

ahmed-irfan (changes requested):
Left some comments

src/utils.cpp
makaimann: So I think it will be removed below because it by definition is not in the unsat core (since it wasn't assumed in this iteration). Although, one minor thing to note is that `vector::erase` is not very efficient. It might even be better to clear the vector and just add things back into it.

ahmed-irfan: Right.

ahmed-irfan (approved):
LGTM. Thank for adding this

PR #154, opened 2021-01-08
Generic Term Header

yoni206:
Header file for generic terms + an empty test that includes it, so that it is compiled.
Implementations are left for future PR(s).


makaimann (approved):
LGTM, thanks!

PR #155, opened 2021-01-08
Generic Term: Implementations and some tests

yoni206:
This PR provides implementations for GenericTerm.
It also provides some testing functions, though the majority of testing for this class will be given via the future `GenericSolver` testing.

### DIFF
--- a/src/generic_term.cpp
+++ b/src/generic_term.cpp
@@ -28,7 +28,6 @@ namespace smt {
 GenericTerm::GenericTerm(Sort s, Op o, TermVec c, string r)
     : sort(s), op(o), children(c), repr(r), is_sym(false), is_par(false)
 {
-  ground = compute_ground();
 }
 
 GenericTerm::GenericTerm(Sort s, Op o, TermVec c, string r, bool is_sym)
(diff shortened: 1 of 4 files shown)


src/generic_term.cpp
```diff
@@ -0,0 +1,196 @@
...
+  return repr == gt->to_string();
+}
+
+string GenericTerm::compute_string() const
+{
+  if (!repr.empty())
+  {
+    return repr;
+  }
+  Assert(!op.is_null());
```
makaimann, line 99: Just checking, does this work because symbols always have their `repr` set on construction?

src/generic_term.cpp
```diff
@@ -0,0 +1,196 @@
...
+      op(o),
+      children(c),
+      repr(r),
+      is_sym(is_sym),
+      is_par(!is_sym),
+      ground(true)
+{
+}
+
+bool GenericTerm::compute_ground()
```
makaimann, line 45: Remind me again, why do we need to know if the term is `ground`?

makaimann (approved):
Looks good to me! Just left a few minor questions for my own benefit.

src/generic_term.cpp
yoni206: yes, that's the idea. although i agree it is not super clear from the code right now. I am open for suggestion if there is a cleaner way for this.

src/generic_term.cpp
yoni206: Good point! I will be using this in the generic solver, but for now it is not needed. So I removed it and will add when it is really needed (couple of PRs from now).

PR #157, opened 2021-01-12
Don't require boolean constants in check-sat-assuming

makaimann:
This PR just makes it so that there's no artificial restriction on the assumptions to `check_sat_assuming`. MathSAT is the only solver that requires they be (negated) boolean constants, so that's the only solver that will throw an exception.

This PR makes other solvers more consistent. E.g. boolector will do top-level propagation, so if you assume:
`lbl <-> formula` 
at the top-level then `lbl` would not be considered a constant anymore, and before this change it would throw an exception when using `lbl` in `check_sat_assuming`. It makes more sense to allow `boolector` to have arbitrary boolean assumptions.

From the SMT-LIB level, the argument that this should be allowed in the API is that you can call `check_sat_assuming` with a `define-fun`. I.e. it's a symbol, but not necessarily a boolean constant. Therefore in the API it makes sense to be able to assume any formula. In the future, we could consider hiding the mathsat restriction at the smt-switch level.

### DIFF
--- a/btor/src/boolector_solver.cpp
+++ b/btor/src/boolector_solver.cpp
@@ -319,28 +319,7 @@ Result BoolectorSolver::check_sat_assuming(const TermVec & assumptions)
   for (auto a : assumptions)
   {
     bt = std::static_pointer_cast<BoolectorTerm>(a);
-
-    bool is_literal = true;
-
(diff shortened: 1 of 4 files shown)


ahmed-irfan (approved):
LGTM

PR #158, opened 2021-01-12
Removing helper functions from generic solver header

yoni206:
In order to make the following PRs shorter, I am removing all helper functions from the generic solver's header.
This way, the helper functions, as well as their definitions, will be added gradually as needed, which will hopefully make the next PRs easier to review.


makaimann (approved):
Looks good!

PR #159, opened 2021-01-13
Generic solver: empty implementations

yoni206:
This PR adds a cpp file for the generic solver. Most functions are empty, except for the constructor and destructor. Also, two helper functions are added to the header, without implementations, as they are called from the constructor and destructor (though currently the call is commented out until they are implemented). The test is extended to call the constructor. Finally, the constructor requires an enum, and so we add a GENERIC_SOLVER enum.


src/solver_enums.cpp
```diff
@@ -51,6 +51,17 @@ const unordered_map<SolverEnum, unordered_set<SolverAttribute>>
               QUANTIFIERS
           } },
 
+        { GENERIC_SOLVER,
+          {
+              TERMITER,
+              THEORY_INT,
```
makaimann, line 57: So what did we decide about this again? What if the generic solver is instantiated with a solver that doesn't support integers? Is this actually just the CVC4 supported attributes?

makaimann (approved):
Left a minor comment but otherwise looks good. Thanks for the PRs!

src/solver_enums.cpp
yoni206: These are the attributes that the generic *interface* supports. It can handle, e.g., integers, but it cannot handle term transfer. So this is a property of the generic interface, not of any particular instantiation of it (which would be impossible to predict). Does that make sense?

makaimann: I see, makes sense!

makaimann:
Looks like there's an issue building it on mac: https://travis-ci.org/github/makaimann/smt-switch/jobs/754242758

yoni206:
Thanks for the heads up! lets see if this will solve it: 119624c

PR #160, opened 2021-01-15
disabling the generic solver on macos

yoni206:
In case smt-switch is exited unexpectedly, we notify the generic solver sub-process.
We do so using a `prctl` system call, that is only available no unix, but not on mac os.
There are mac os alternatives, but we leave the task of using them for future PRs, after the entire generic solver is merged and is working on linux.

This PR disables the generic solver on macos using macros.
In addition, we remove the `SolverEnum` argument from the constructor of the generic solver, as it is known to be `GENERIC_SOLVER`.

Details about prctl and its macos alternatives:
https://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits

### DIFF
--- a/include/generic_solver.h
+++ b/include/generic_solver.h
@@ -31,8 +31,7 @@ class GenericSolver : public AbsSmtSolver
   GenericSolver(std::string path,
                 std::vector<std::string> cmd_line_args,
                 uint write_buf_size = 256,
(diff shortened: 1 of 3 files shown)


makaimann (approved):
LGTM, thanks!

PR #161, opened 2021-01-19
Communicating with Generic Solvers

yoni206:
This PR introduces the code that communicates with a solver binary, runs commands, and read their results.
Only the infrastructure for that is included, as well as support for `set option` command.
More commands will be added in subsequent PRs.

The generic solver test now tries to communicate with all available solver binaries.

### DIFF
--- a/src/generic_solver.cpp
+++ b/src/generic_solver.cpp
@@ -48,7 +48,7 @@ namespace smt {
 
 // helper functions
 //
-bool new_line(char c) { return (c == '\n' || c == '\r' || c == 0); }
+bool is_new_line(char c) { return (c == '\n' || c == '\r' || c == 0); }
 
 // from: https://stackoverflow.com/a/36000453/1364765
 std::string & trim(std::string & str)
@@ -197,7 +197,7 @@ bool GenericSolver::is_done(int just_read, std::string result) const
         count--;
       }
     }
-    done = (count == 0) && new_line(result[result.size() - 1]);
+    done = (count == 0) && is_new_line(result[result.size() - 1]);
   }
   else
   {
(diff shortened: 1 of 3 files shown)


tests/test-generic-solver.cpp
```diff
@@ -43,11 +43,58 @@
...
+  path += "/build/bin/yices_smt2";
+  args = { "--incremental" };
+  test_binary(path, args);
+#endif
+
+  // testing with cvc4 binary
+#if BUILD_BTOR
+  path = (STRFY(BTOR_HOME));
+  path += "/build/bin/btor";
+  args = { "--incremental" } test_binary(path, args);
```
makaimann, line 97: Missing semi-colon.

src/generic_solver.cpp
```diff
@@ -46,6 +46,27 @@ using namespace std;
 
 namespace smt {
 
+// helper functions
+//
+bool new_line(char c) { return (c == '\n' || c == '\r' || c == 0); }
```
makaimann, line 51: suggestion: `is_new_line`

src/generic_solver.cpp
```diff
@@ -81,16 +101,189 @@ GenericSolver::~GenericSolver() {
...
+    written_chars += substr_size;
+  }
+}
+
+bool GenericSolver::is_done(int just_read, std::string result) const
+{
+  bool done = false;
+  int count = 0;
+  // if we didn't read anything now, the command is done executing
+  if (just_read == 0)
```
makaimann, line 181: This is fine but a little funny to me that there's a flag that can determine the output. Do you only need `just_read` for this and the assertion below? I think it's fine as-is just thought I'd bring it up for you to consider.

makaimann (approved):
Looks really good to me! Left a few minor comments, one of which should fix the test errors.

PR #162, opened 2021-01-20
Generic solver implementations: creating sorts

yoni206:
This PR provides the implementations of the various `make_sort` functions.
It also extends the test file, where each of the available binaries is tested only on sorts that it supports.


makaimann (approved):
LGTM!

makaimann:
Although the Travis build failed, that was due to the strange issue downloading ANTLR for CVC4. I'm going to merge, and if it still fails we can always fix it later.

PR #164, opened 2021-01-21
Temporary fix: download pre-compiled CVC4

makaimann:
There appears to be an issue using the [get-anltr-3.4](https://github.com/CVC4/CVC4/blob/master/contrib/get-antlr-3.4) script from CVC4 on the Travis Mac environment -- it keeps failing with "Connection refused". It is not clear why, but a temporary fix is to just download a pre-compiled CVC4 instead of building it in Travis. This also has the nice effect of speeding up the build.


makaimann:
Looks like the issue was resolved somehow! Must have been something on the Travis end.

PR #165, opened 2021-01-21
Generic Solver: basic commands, basic terms

yoni206:
This PR introduces support for `push`, `pop`, `check-sat`, `assert`, and `declare-fun`.
A new test is added to test all these commands. Since no `make_term` functions were implemented so far, the test only involves Boolean variables.

Unfortunately, the PR shows a bigger diff than there is in reality due to some `clang-format`s, white spaces, and a minor cosmetic refactoring I did to the test file.

### DIFF
--- a/src/generic_solver.cpp
+++ b/src/generic_solver.cpp
@@ -18,29 +18,28 @@
 ** https://stackoverflow.com/a/6172578/1364765
 */
 
-
 // generic solvers are not supported on macos
 #ifndef __APPLE__
 
-#include "assert.h"
-
 #include "generic_solver.h"
-#include "sort_inference.h"
-#include "sort.h"
-#include "smtlib_utils.h"
-#include "utils.h"
 
-
-#include<unistd.h>
-#include<sys/wait.h>
-#include<sys/prctl.h>
-#include<signal.h>
-#include<stdlib.h>
-#include<string.h>
-#include<stdio.h>
-#include<algorithm>
-#include<fcntl.h>
+#include <fcntl.h>
 #include <poll.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/prctl.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include <algorithm>
+
+#include "assert.h"
+#include "smtlib_utils.h"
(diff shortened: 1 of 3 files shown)


src/generic_solver.cpp
```diff
@@ -66,12 +65,19 @@ std::string & trim(std::string & str)
 }
 
 // class methods implementation
-GenericSolver::GenericSolver(string path, vector<string> cmd_line_args, uint write_buf_size, uint read_buf_size)
-  : AbsSmtSolver(SolverEnum::GENERIC_SOLVER), path(path), cmd_line_args(cmd_line_args), write_buf_size(write_buf_size), read_buf_size(read_buf_size),
-  name_sort_map(new unordered_map<string, Sort>()),
-  sort_name_map(new unordered_map<Sort, string>()),
-  name_term_map(new unordered_map<string, Term>()),
-  term_name_map(new unordered_map<Term, string>())
+GenericSolver::GenericSolver(string path,
```
makaimann, line 68: Not sure if this is just github but are there a bunch of trailing spaces on this line?

src/generic_solver.cpp
```diff
@@ -66,12 +65,19 @@ std::string & trim(std::string & str)
...
+                             vector<string> cmd_line_args,
+                             uint write_buf_size,
+                             uint read_buf_size)
+    : AbsSmtSolver(SolverEnum::GENERIC_SOLVER),
+      path(path),
+      cmd_line_args(cmd_line_args),
+      write_buf_size(write_buf_size),
+      read_buf_size(read_buf_size),
+      name_sort_map(new unordered_map<string, Sort>()),
```
makaimann, line 77: Are these deleted somewhere? I don't see it.

makaimann (commented):
Hey, looks pretty good! Just had a couple questions.

src/generic_solver.cpp
yoni206: I can't see anything :) And this specific diff is due to clang-format.

src/generic_solver.cpp
makaimann: Talked off line -- I missed that this memory is managed by a `unique_ptr`

makaimann (approved):
LGTM!

PR #167, opened 2021-01-22
Generic Solver: Infrastructure for terms + Boolean constants

yoni206:
This PR adds the infrastructure to create terms via the generic solver.

Most of the functions that are added are `protected`, as they deal with the internal state of the generic solver and how it communicates with the binary.

The only new public function that is added allows to create Boolean constants.

The test file is extended with a test that involves Boolean constants.

### DIFF
--- a/src/generic_solver.cpp
+++ b/src/generic_solver.cpp
@@ -566,7 +566,7 @@ Term GenericSolver::store_term(Term term) const
     // we send a command:
     // `(define-fun t1 () Int 0)`
     // If in the future we see a term `0+0`, it will
-    // be send to the solver as `(+ t1 t1)`.
+    // be sent to the solver as `(+ t1 t1)`.
     //
     // However, terms with bound variables cannot be given to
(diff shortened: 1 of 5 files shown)


src/generic_solver.cpp
```diff
@@ -485,9 +542,60 @@ Term GenericSolver::get_selector(const Sort & s, std::string con, std::string na
...
+    string name;
+    // ground terms are communicated to the binary
+    // using a define-fun command.
+    // In future instances of this term, we will use the name
+    // of the defined function.
+    // For example: if `term` is `0`, and `name` is `t1`,
+    // we send a command:
+    // `(define-fun t1 () Int 0)`
+    // If in the future we see a term `0+0`, it will
+    // be send to the solver as `(+ t1 t1)`.
```
makaimann, line 569: Very minor: "sent"

makaimann (approved):
Looks great! Left a very minor comment.

PR #170, opened 2021-01-25
Generic Solver: get-value, get-unsat-core, check-sat-assuming

yoni206:
This PR introduces implementations of `get-value`, `get-unsat-core` and `check-sat-assuming` in the generic solver.
It also removes support for non 0-ary uninterpreted sorts (the implementation was added by mistake previously).
Tests are added for the new functions.

### DIFF
--- a/src/generic_solver.cpp
+++ b/src/generic_solver.cpp
@@ -349,40 +349,7 @@ std::string GenericSolver::to_smtlib_def(Term term) const
 }
 
 Sort GenericSolver::make_sort(const Sort & sort_con, const SortVec & sorts) const {
-  // When constructing a sort, the sort constructor
-  // must have been already processed
-  assert(sort_name_map->find(sort_con) != sort_name_map->end());
-  // The arity of the sort constructor must match
-  // the number of arguments
-  assert(sort_con->get_arity() == sorts.size());
-
-  // All the input sorts must have been processed
-  for (Sort sort : sorts)
-  {
-    assert(sort_name_map->find(sort) != sort_name_map->end());
-  }
-
-  // creating the new sort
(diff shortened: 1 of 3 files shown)


src/generic_solver.cpp
```diff
@@ -349,40 +349,7 @@ std::string GenericSolver::to_smtlib_def(Term term) const
...
-  {
-    return (*name_sort_map)[name];
-  }
-  else
-  {
-    (*name_sort_map)[name] = sort;
-    (*sort_name_map)[sort] = name;
-    return sort;
-  }
+  assert(false);
```
makaimann, line 352: Is this going to be implemented later or not at all? If it's the latter could you throw an exception instead?

src/generic_solver.cpp
```diff
@@ -868,7 +1007,29 @@ Result GenericSolver::check_sat()
 
 Result GenericSolver::check_sat_assuming(const TermVec & assumptions)
 {
-  assert(false);
+  string names;
+  for (Term t : assumptions)
+  {
+    // assumptions can only be Boolean literals
+    if (!t->is_symbolic_const() || t->get_sort()->get_sort_kind() != BOOL)
```
makaimann, line 1014: Just FYI, we relaxed this requirement recently. See the reasoning here: https://github.com/makaimann/smt-switch/pull/157. Basically the TLDR is that through the SMT-LIB interface you can use a `define-fun` without creating a fresh literal as your assumption, and so through the API it makes sense to allow any boolean formula. Only mathsat has the restriction that it needs to be a boolean literal (as far as I know).

makaimann (approved):
Looks good to me! Just left a few minor comments.

PR #171, opened 2021-01-26
Provide list and set versions of check_sat_assuming

makaimann:
Adds basic support for a `check_sat_assuming` method that takes a list. Because of ambiguity with brace-enclosed lists, this function needs to have a different name. Furthermore, it would be nice if the code could be the same for both cases, but this requires templates and that doesn't work with virtual functions. For now this should work and it can be improved in the future.

The use case is for quick iterations of `check_sat_assuming_list` calls because it is faster to remove elements from a list.

### DIFF
--- a/btor/include/boolector_solver.h
+++ b/btor/include/boolector_solver.h
@@ -58,6 +58,7 @@ class BoolectorSolver : public AbsSmtSolver
   Result check_sat() override;
   Result check_sat_assuming(const TermVec & assumptions) override;
   Result check_sat_assuming_list(const TermList & assumptions) override;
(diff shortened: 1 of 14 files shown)


makaimann:
@zhanghongce here's the PR. Does this look like it would work for you?

makaimann:
FYI @yoni206 and @lstuntz for this interface addition. You shouldn't have to do anything because it just throws an exception by default which is fine. I plan to come back later and look into cleaner solutions (it will likely be some kind of messy template specialization).

zhanghongce (commented):
Thank you @makaimann

PR #173, opened 2021-01-29
Generic Solver: Supporting and testing various buffer sizes

yoni206:
So far all buffer sizes were supported, but only one was tested (buffer size of 5).
In this PR we introduce a bound for buffer sizes, between 2 to 256 (inclusive),
and test a representative subset of the possible values.
The reason for this particular bound is that when the buffer size was 1 or > 1000 the communication with the binary of the solver froze. 256 is a bit conservative but should be enough for now.


makaimann (approved):
Looks great to me, thanks!

PR #174, opened 2021-02-03
Clearly documented what is supported by generic solvers

yoni206:
This PR clearly documents the unsupported parts of the generic solver.
It also throws `NonImplementedException`s rather than `assert(false)` when these functions are called.


makaimann (approved):
LGTM, thanks for the documentation!

PR #175, opened 2021-02-04
Generic Solver: Refactoring + bugfix

yoni206:
This PR refactors the creation of values, by separating it to two parts:
1. Creation of a GenericTerm object that represents a value.
2. Sending a define-fun command to the solver's binary with the created value.

This actually solves a bug: when creating a Term from the result of a `get-value`, we used to send a `define-fun` command to the solver with that value. This took the solver out of `sat` state, and then we couldn't perform another `get-value`. This is now fixed, and the test file is extended to test this scenario.

In addition, we now support get-value for Booleans, which we should. This is also tested by the same test now.

### DIFF
--- a/include/generic_solver.h
+++ b/include/generic_solver.h
@@ -117,6 +117,21 @@ class GenericSolver : public AbsSmtSolver
   /******************
    * helper methods *
    *******************/
+
+  /** a helper function for the corresponding make_term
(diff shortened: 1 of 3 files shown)


makaimann (approved):
Looks good, thanks!

PR #176, opened 2021-02-05
checking errors and fixes

yoni206:
1. checking that the result from the solver does not indicate an error, and throwing an exception otherwise.
2. adding support for `print_value_as`
3. changing the string representation of `Int_To_BV` to conform with the way `CVC4` parses it.

Tests are not added, because these features are not consistent between solvers. All three changes will be tested in a future PR, when we add a CVC4-based generic solver to the available solvers for testing.

### DIFF
--- a/src/ops.cpp
+++ b/src/ops.cpp
@@ -84,7 +84,7 @@ const std::unordered_map<PrimOp, std::string> primop2str(
       { Rotate_Left, "rotate_left" },
       { Rotate_Right, "rotate_right" },
       { BV_To_Nat, "bv2nat" },
-      { Int_To_BV, "nat2bv" },
+      { Int_To_BV, "int2bv" },
       { Select, "select" },
       { Store, "store" },
       { Forall, "forall" },
--- a/include/generic_solver.h
+++ b/include/generic_solver.h
@@ -131,6 +131,15 @@ class GenericSolver : public AbsSmtSolver
   // returns a string representation of a term in smtlib
   std::string to_smtlib_def(Term term) const;
 
+  // when an SMT-LIB compliant solver is supposed
+  // to return a result (e.g., get-value),
+  // a result that starts with "(error " indicates
+  // that an error occurred.
+  // This cannot be caught by print-success,
+  // which is not utilized for commands that
+  // expect a result.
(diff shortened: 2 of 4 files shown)


src/ops.cpp
```diff
@@ -84,7 +84,7 @@ const std::unordered_map<PrimOp, std::string> primop2str(
       { Rotate_Left, "rotate_left" },
       { Rotate_Right, "rotate_right" },
       { BV_To_Nat, "bv2nat" },
-      { Int_To_BV, "nat2bv" },
+      { Int_To_BV, "int2bv" },
```
makaimann, line 87: Is it actually `int2bv` in SMT-LIB? I thought `nat2bv` was right.

makaimann (approved):
Looks good, thanks! Just left a minor question.

src/ops.cpp
yoni206: Well we are testing the generic solver with CVC4, which expects `int2bv`...

makaimann: Gotcha, that's a little strange since it's `nat2bv` here: http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml. Doesn't seem like there's an easy workaround for now. But maybe just something to keep in mind in case it becomes a problem later.

yoni206: well -- in the link that you sent this operator is not defined in the theory itself, but as a helper to define the semantics. So I am not sure it is officially part of the theory.

In partice, i think this can only occur in logics that combine integer arithmetic (linear or not linear) with BV. There are only two such logics in SMT-LIB, and non of their benchmarks include this.

```
SMT-LIB/non-incremental$ ls  |grep BV |grep I
AUFBVDTLIA
AUFBVDTNIA
SMT-LIB/non-incremental$ grep nat2bv -rI AUFBVDTLIA AUFBVDTNIA
SMT-LIB/non-incremental$
```

Maybe this means we should ban this when using a generic solver?

makaimann: Ahh I see, my bad. I think if it's not standardized it's fair to just pick one and go with it. In this case, CVC4 makes the most sense so this is perfect.

PR #177, opened 2021-02-09
Adding the generic solver to the test suite

yoni206:
This PR adds the generic solver to the solvers that are used for testing.
It only use it with the CVC4 binary for these tests.

This required excluding the generic solver from some tests, making some local case splits inside some tests, and also changing the body of some tests in a minor way.

### DIFF
--- a/tests/unit/unit-solving-interface.cpp
+++ b/tests/unit/unit-solving-interface.cpp
@@ -36,7 +36,6 @@ class UnitSolveTests : public ::testing::Test,
     s = create_solver(GetParam());
     s->set_opt("incremental", "true");
     s->set_opt("produce-models", "true");
-
     boolsort = s->make_sort(BOOL);
     bvsort = s->make_sort(BV, 4);
   }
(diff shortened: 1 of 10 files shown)


tests/unit/unit-solving-interface.cpp
```diff
@@ -66,16 +66,22 @@ TEST_P(UnitSolveTests, CheckSatAssuming)
     // mathsat is the only solver (so far)
     // to have the restriction that assumptions must be
     // (negated) boolean constants
-    EXPECT_EQ(s->get_solver_enum(), MSAT);
+    // Generic solvers are SMT-LIB compliant, and threfore
```
makaimann, line 69:
```suggestion
    // Generic solvers are SMT-LIB compliant, and therefore
```

makaimann: quick question -- I thought they wouldn't actually need to follow this restriction because you use define-funs and those are allowed in `check-sat-assuming`. Is that wrong?

makaimann (approved):
Looks good to me! All things considered, it's not too bad in terms of case-splits / changing the parameterizations. Pretty cool! Just left a few minor comments. Also, I plan to fix the `produce-unsat-assumptions` mismatch soon. That will require accommodating changes in tools that depend on smt-switch.

tests/unit/unit-solving-interface.cpp
yoni206: huh, interesting, let me check.

PR #180, opened 2021-02-26
Add Z3 functionality (terms and sorts)

lstuntz:
Amalee and Makai have reviewed sub-prs for this
sorts: https://github.com/lstuntz/smt-switch/pull/4 (reviewed)
terms: https://github.com/lstuntz/smt-switch/pull/5 (reviewed)
formatting: https://github.com/lstuntz/smt-switch/pull/6 (not reviewed, just formatting/reorganizing)


configure.sh
```diff
@@ -16,10 +16,12 @@ Configures the CMAKE build environment.
 --yices2                build yices2            (default: off)
 --cvc4                  build cvc4              (default: off)
 --msat                  build MathSAT           (default: off)
+--z3					build z3				(default: off)
```
makaimann, line 19: minor: line up "build" with other rows

configure.sh
```diff
@@ -16,10 +16,12 @@ Configures the CMAKE build environment.
...
 --btor-home=STR         custom BTOR location    (default: deps/boolector)
 --yices2-home=STR       custom YICES2 location  (default: deps/yices2)
 --cvc4-home=STR         custom CVC4 location    (default: deps/CVC4)
 --msat-home=STR         custom MathSAT location (default: deps/mathsat)
+--z3-home=STR			custom Z3 location 		(default: deps/z3)
```
makaimann, line 24: minor: line up "custom"

z3/src/z3_solver.cpp
```diff
@@ -0,0 +1,844 @@
...
+
+  Z3_ast c_array = Z3_mk_const_array(ctx, zsort->type, zterm->term);
+  expr final = to_expr(ctx, c_array);
+  return std::make_shared<Z3Term>(final, ctx);
+}
+
+void Z3Solver::assert_formula(const Term & t)
+{
+  throw NotImplementedException(
+      "Term iteration not implemented for Z3 backend.");
```
makaimann, line 238: Just FYI, looks like this exception message was copy-pasted. This one doesn't need to be fixed now since it will be implemented soon. But for anything that stays throwing an exception permanently, please update the message.

makaimann (approved):
Just made a final pass. Looks good to me! Left a few minor comments that could be fixed before merging but are not crucial.

configure.sh
lstuntz: So strangely enough this is lined up on my local master but isn't registering as a change that I should add/commit. Thoughts on how to fix that?

configure.sh
lstuntz: (same problem as above)

configure.sh
makaimann: Oh then it's probably a tabs vs spaces issue. If you show whitespace characters, there are probably tabs there which you can replace with spaces to make sure it's lined up on other editors. https://stackoverflow.com/questions/6255839/show-space-tab-crlf-characters-in-editor-of-visual-studio

lstuntz:
(for the sake of consistency with my original comment)
assert_formula and check: https://github.com/lstuntz/smt-switch/pull/7 (not yet reviewed, added to this pr upon request)

configure.sh
amaleewilson: You can also tell your editor to always do spaces when you hit tab. 
https://stackoverflow.com/questions/29972396/how-can-i-customize-the-tab-to-space-conversion-factor

And there is an extension to replace existing tabs with spaces.
https://marketplace.visualstudio.com/items?itemName=johnnywong.vscode-ts-whitespace-converter

lstuntz: thanks! (They were tabs, whereas all the others were spaces!)

PR #182, opened 2021-03-01
Adding Z3 to test infrastructure, but excluding from all tests

yoni206:
This PR adds Z3 to the test infrastructure.

Since some tests are failing, the actual inclusion of Z3 in the tests is commented out.
From a sample of failures I looked at, it seems that some more implementations are needed (e.g., `set_opt`, `get_op`). 

To include Z3 in the tests, uncomment [this line](https://github.com/makaimann/smt-switch/pull/182/files#diff-e684bb524326e70efef455413e1e99e49c5403c1c9f346297e7a12a281375457R75).

To investigate a specific test, run:
```
cd build
make
./tests/./<test-name> (e.g., ./tests/./unit-op)
```

@lstuntz , FYI. Also, I am happy to walk you through the process of running a test and fixing it!


makaimann (approved):
Looks great to me, thanks!

PR #185, opened 2021-03-02
Add portfolio solving.

amaleewilson:
This is a first pass and under-tested. 

Am I doing the term translation right? 

Need to look into whether detach()'s and return is leaking resources (probably is).


include/portfolio_solver.h
```diff
@@ -0,0 +1,61 @@
+
+#pragma once
+
+#include "smt_defs.h"
```
amaleewilson, line 4: This includes everything right now because of a bunch of copying and pasting. Will remove unnecessary includes.

include/portfolio_solver.h
```diff
@@ -0,0 +1,61 @@
...
+
+  bool portfolio_solve(SmtSolver og, std::vector<SmtSolver> solvers, Term t) {
+
+    for (auto s : solvers) {
+      std::thread t1(do_solver, s, t);
+      t1.detach();
+    }
+
+    // while !someone_done, wait
+    // once someone is done, signal exit and return
```
amaleewilson, line 54: This thread does not signal an exit.

tests/CMakeLists.txt
```diff
@@ -148,6 +148,11 @@ target_link_libraries(test_generic_sort gtest_main)
 target_link_libraries(test_generic_sort test-deps)
 add_test(NAME test_generic_sort_test COMMAND test_generic_sort)
 
+add_executable(test_portfolio_solver "${PROJECT_SOURCE_DIR}/tests/test-portfolio-solver.cpp")
+target_link_libraries(test_portfolio_solver gtest_main)
+target_link_libraries(test_portfolio_solver test-deps)
+add_test(NAME test_portfolio_solver_test COMMAND test_portfolio_solver)
+
```
amaleewilson, line 155: Is this an okay place for this test?

tests/test-portfolio-solver.cpp
```diff
@@ -0,0 +1,77 @@
...
+  Term test_term = s->make_term(Not,
+                   s->make_term(Implies,
+                                s->make_term(Equal, x, y),
+                                s->make_term(Equal,
+                                             s->make_term(Select, arr, x),
+                                             s->make_term(Select, arr, y))));
+
+  SmtSolver s1 = MsatSolverFactory::create(false);
+  SmtSolver s2 = MsatSolverFactory::create(false);
+  SmtSolver s3 = MsatSolverFactory::create(false);
```
amaleewilson, line 61: Currently testing with several MathSat solvers and about to try adding Boolector and CVC4.

include/portfolio_solver.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace smt {
+  std::mutex m;
+  std::condition_variable cv;
+  bool someone_done = false;
+  bool is_sat = false;
+
+  bool do_solver(SmtSolver s, Term t) {
+    TermTranslator to_s1(s);
+    Term a = to_s1.transfer_term(t);
```
amaleewilson, line 39: Is this correct term translation?

include/portfolio_solver.h
```diff
@@ -0,0 +1,61 @@
...
+    s->assert_formula(a);
+    is_sat = s->check_sat().is_sat();
+    someone_done = true;
```
amaleewilson, line 42: Probably needs a lock here, too. Will update it.

makaimann:
> Am I doing the term translation right?

Yep, you are using term translation right! The only thing I'd recommend is doing `transfer_term(Term, smt::BOOL)`. This to handle an edge case. For example, if the term you're checking was created by a non-logging BoolectorSolver, then it will treat all booleans as bitvectors of width 1. If you translate that to another solver it will complain. Passing `BOOL` tells it to interpret the term as a boolean (if possible).

Thanks for working on this! It will be very cool to have.

amaleewilson:
Ready for review. There's only one test right now, but I could add more if it's useful.

include/portfolio_solver.h
```diff
@@ -0,0 +1,85 @@
...
+  std::unique_lock<std::mutex> lk(m);
+  cv.wait(lk, [] { return a_solver_is_done; });
+  for (int i = 0; i < pthreads.size(); ++i)
+  {
+    pthread_cancel(pthreads[i]);
+  }
+
+  return is_sat;
+}
+}  // namespace smt
```
ahmed-irfan, line 85: new line

include/portfolio_solver.h
```diff
@@ -0,0 +1,85 @@
...
+#include <condition_variable>
+#include <mutex>
+#include <thread>
+
+#include "smt.h"
+
+namespace smt {
+
+// Used for synchronization.
+std::mutex m;
```
ahmed-irfan, line 27: how about moving the implementation in a separate cpp file? Doing so, the global variables will moved to the cpp file .

tests/CMakeLists.txt
```diff
@@ -50,6 +50,10 @@ if (BUILD_YICES2)
   add_subdirectory(yices2)
 endif()
 
+if (${BUILD_YICES2} AND ${BUILD_BTOR} AND ${BUILD_MSAT} AND ${BUILD_CVC4})
```
ahmed-irfan, line 53: we will build portfolio only if all the solvers are available. Is it intended?

ahmed-irfan (commented):
Quick review, I didn't go into the threading part yet.

tests/CMakeLists.txt
amaleewilson: This is building the portfolio solver test that uses all four solvers.

include/portfolio_solver.h
amaleewilson: Not sure what you mean. You want a new line between 84 and 85?

include/portfolio_solver.h
amaleewilson: I'm not super opposed to separating it into a header/implementation file. My hesitation was that it's super tiny and not a class, so why split it?

tests/CMakeLists.txt
ahmed-irfan: oh, i missed that. then it is fine

include/portfolio_solver.h
ahmed-irfan: add a line break at the end of 85.

include/portfolio_solver.h
ahmed-irfan: the problem is that when we include this header file, those global variables will be visible, which I think is not intended

amaleewilson: That's the best point

CMakeLists.txt
```diff
@@ -19,25 +19,26 @@ if (NOT SMT_SWITCH_LIB_TYPE)
 endif()
 
 add_library(smt-switch "${SMT_SWITCH_LIB_TYPE}"
```
amaleewilson, line 21: Needed to add portfolio_solve here. Hope it's okay that I alphabetized the list while I was at it.

makaimann: Not at all, that's great!

include/portfolio_solve.h
```diff
@@ -0,0 +1,39 @@
...
+ *  @param t The term being translated to solver s.
+ */
+void run_solver(SmtSolver s, Term t);
+
+/** Launch many solvers and return whether the term is satisfiable when one of
+ *  them has finished.
+ *  @param solvers The solvers to run.
+ *  @param t The term to be checked.
+ */
+bool portfolio_solve(std::vector<SmtSolver> solvers, Term t);
```
makaimann, line 37: What do you think about returning an `smt::Result` instead? Then it will fit into infrastructure using individual solvers better. Plus, you might actually want three possible outputs. E.g., if all solvers return unknown then you should return unknown.

src/portfolio_solve.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace smt {
+
+// Used for synchronization.
+std::mutex m;
+std::condition_variable cv;
+
+// Once a solver is done, is_sat has been set,
+// and the main thread can terminate the others.
+bool a_solver_is_done = false;
```
makaimann, line 27: This might just be me not understanding the multi-threading part, but why is this variable outside the function? What if you call `portfolio_solve` twice in a row? Is it reset when `portfolio_solve` completes?

tests/portfolio/test-portfolio-solver.cpp
```diff
@@ -0,0 +1,93 @@
+/*********************                                                        */
+/*! \file test-generic-solver.cpp
```
makaimann, line 2: minor: file name.

makaimann (commented):
It looks good -- thanks for working on this! I left a few comments, mostly minor, and one that might just be me misunderstanding.

include/portfolio_solve.h
amaleewilson: It should definitely return an smt::Result.

src/portfolio_solve.cpp
amaleewilson: I failed to imagine a case where someone would call portfolio_solve twice in a row, but I can make it so that it handles that. 

The variable is outside the function because all threads that execute the function `do_solver` (and the main thread executing `portfolio_solve`) are sharing this variable. Once a thread has finished, it acquires the mutex and marks a_solver_is_done as true. The condition variable in portfolio_solve is waiting for a_solver_is_done to be true. An alternative way to write this is 

```
while (!a_solver_is_done) {} // poll until true
// now cancel threads executing do_solver and return
```

makaimann: Gotcha, makes sense! I think it would be good to support multiple calls in a row if it's not too difficult. And if it is too difficult, it should throw an exception the second call. It's probably a pretty common use case to call multiple times. Even if it's not solving "incremental solving" directly, it could still be used within another framework (symbolic execution or a modeling tool) which might make many separate satisfiability checking calls in the same process.

tests/portfolio/test-portfolio-solver.cpp
```diff
@@ -0,0 +1,93 @@
...
+  SmtSolver s2 = MsatSolverFactory::create(false);
+  SmtSolver s3 = MsatSolverFactory::create(false);
+  SmtSolver s4 = BoolectorSolverFactory::create(false);
+  SmtSolver s5 = BoolectorSolverFactory::create(false);
+  SmtSolver s6 = Yices2SolverFactory::create(true);
+  SmtSolver s7 = CVC4SolverFactory::create(false);
+  SmtSolver s8 = CVC4SolverFactory::create(false);
+  SmtSolver s9 = Yices2SolverFactory::create(true);
+  vector<SmtSolver> solvers;
+  s1->set_opt("produce-models", "true");
```
ahmed-irfan, line 75: do we need this option?

amaleewilson:
Okay, y'all, so here's where things stand right now: 

Killing a thread while it is working is hard to do without getting 

```
terminate called without an active exception
Aborted
```

The initial version that was committed did not successfully cancel the threads executing the run_solver function, and the current version (both pushed and unpushed) is also failing to do this. 

First off, I'm fairly certain that calling detach on a std:thread has some unseen effect on the native handle (which I was trying to use to cancel it). 

The second issue that is more difficult to overcome is that default pthreads_cancel will wait for a "cancellation point" (examples of cancellation points can be found here: https://man7.org/linux/man-pages/man7/pthreads.7.html). Unfortunately, there do not appear to be any suitable cancellation points while a solver is running. I tried two things: asynchronous cancellation and inserting a cancellation point. Both failed. 

I tried setting the pthreads cancel type to asychronous (bad idea because the program state won't be well defined), and the threads do stop, but terminate gets called, and we see the terminate error above. 

For the cancellation point, I put a sleep right after the call to s->check_sat so that reset could be called on the PortfolioSolver, but that results in nondeterministic exceptions/segfaults. 

So I don't think we can trust pthread_cancel to behave as we might expect/wish it to, and because of the untrustworthiness of thread cancellation, this is implemented as a class (instead of a function that resets its variables). Alternatively, we could insert a thread.join(), but that would require waiting on all of the solver threads to finish before continuing and seems contrary to the point. 

But I think that the solvers only need one thread anyway, so maybe it's not super egregious that the other solvers have to eventually finish? I can keep looking into this, but I've spent like two days on it, and unless the solvers themselves expose a stop or interrupt, I'm pretty confident that there's not a nice way to shut down all the solvers.

makaimann:
Thanks for looking into this @amaleewilson! Sorry it ended up being such a time sink. It doesn’t need to be perfect and you’ve already done a great job exploring the options, so I wouldn’t recommend spending any more time on this. We can discuss the best approach given these options.

amaleewilson:
So I think the problem with this approach is that we're trying to use threads, but it makes more sense to use processes.

makaimann:
Ahh very interesting. So instead you would use `fork` to create a child process? I haven't used that in C++ at all.

ahmed-irfan:
> Okay, y'all, so here's where things stand right now:
> 
> Killing a thread while it is working is hard to do without getting this error or something similar:
> 
> ```
> terminate called without an active exception
> Aborted
> ```

Sorry for commenting a bit late. 
A thought -- might be silly -- how about having a try-catch block in `run_solver` ?

amaleewilson:
> Ahh very interesting. So instead you would use `fork` to create a child process? I haven't used that in C++ at all.

Yes, or something similar.

tests/portfolio/test-portfolio-solver.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+using namespace smt;
+using namespace std;
+
+int main()
+{
+  cout << "testing portfolio solver" << endl;
+  SmtSolver s = MsatSolverFactory::create(false);
+  s->set_opt("produce-models", "true");
+  Sort intsort = s->make_sort(BV, 10);
```
makaimann, line 32: minor: but I usually follow a naming convention where `intsort` would refer to an integer sort, and for bit vectors I typically use `bvsort` or `bvsort<size>`.

tests/portfolio/test-portfolio-solver.cpp
```diff
@@ -0,0 +1,117 @@
...
+  SmtSolver s2 = MsatSolverFactory::create(false);
+  SmtSolver s3 = MsatSolverFactory::create(false);
+  SmtSolver s4 = BoolectorSolverFactory::create(false);
+  SmtSolver s5 = BoolectorSolverFactory::create(false);
+  SmtSolver s6 = Yices2SolverFactory::create(true);
+  SmtSolver s7 = CVC4SolverFactory::create(false);
+  SmtSolver s8 = CVC4SolverFactory::create(false);
+  SmtSolver s9 = Yices2SolverFactory::create(true);
+  vector<SmtSolver> solvers;
+  solvers.push_back(s6);
```
makaimann, line 75: also minor: doesn't matter but why are they pushed into the vector in a different order?

makaimann (approved):
Looks great, thanks for working on this! I just left a few minor comments.

tests/portfolio/test-portfolio-solver.cpp
amaleewilson: Oops, forgot to fix that when I changed it.

tests/portfolio/test-portfolio-solver.cpp
amaleewilson: They were initially pushed s1 to s9, but I was checking to see if order affected the "winner." I'll put them back in order.

PR #189, opened 2021-03-04
Temporarily disable Z3 build

makaimann:
This PR temporarily disables the Z3 build while the Z3 backend is tested and the final features are being added. Once development is complete, it will be re-enabled.


makaimann:
As discussed offline, this is just temporary so that users don't try the Z3 backend before it is ready.

lstuntz (approved):

makaimann:
Thanks for the review!

PR #191, opened 2021-03-07
Skip test_pysmt if pysmt not installed

makaimann:


makaimann:
Oh thanks for pointing that out! Yeah I'd like to drop that. Don't want `pysmt` as an actual dependency, even for tests. But I'll make sure it gets tested on Travis.

cdonovick:
Also I probably should put a note in the documentation but you can install the extras with:
```bash
$ pip install -e ./python/[pysmt] # install with pysmt
$ pip install -e ./python/[test] # installs with pytest and pysmt (currently)
```
Note in some shells (notably zsh) you need to escape the brackets or quote the path:
```bash
$ pip install -e ./python/\[pysmt\]
$ pip install -e "./python/[test]"
```

cdonovick:
Ill make pr fixing the docs

PR #193, opened 2021-03-07
Update quantifier interface

makaimann:
This PR allows binding several variables at once with a quantifier (as opposed to nesting them). This simplifies the parser implementation and can be more convenient.


btor/src/boolector_solver.cpp
```diff
@@ -935,6 +925,30 @@ Term BoolectorSolver::apply_prim_op(PrimOp op, Term t0, Term t1, Term t2) const
...
+    {
+      std::shared_ptr<BoolectorTerm> bt0 =
+          std::static_pointer_cast<BoolectorTerm>(t0);
+      std::shared_ptr<BoolectorTerm> bt1 =
+          std::static_pointer_cast<BoolectorTerm>(t1);
+      std::shared_ptr<BoolectorTerm> bt2 =
+          std::static_pointer_cast<BoolectorTerm>(t2);
+      std::vector<BoolectorNode *> params({ bt0->node, bt1->node });
+      return std::make_shared<BoolectorTerm>(
+          btor, boolector_forall(btor, &params[0], 2, bt2->node));
```
lstuntz, line 938: You seem to be telling boolector to expect 2 params but then only passing in the first rather than the entire vector

lstuntz (approved):
I think there may be tiny bug with 2 parameter quantifiers for boolector (hopefully I left the comment correctly), but this looks great!

makaimann:
Comment looks great, thanks for the review!

btor/src/boolector_solver.cpp
makaimann: Thanks! It's actually correct, it's just a strange pattern. Since boolector is a C API they don't have vectors, so you can pass the address of the first element (which it treats as an array) and the number of elements. However, someone taught me recently that I can instead use `params.data()` which is much easier to read. I made that change in 496e4533b1a8e0fe1c844281d12f7996db19df43. Thanks for pointing this out!

ahmed-irfan (approved):
LGTM

makaimann:
Thanks for the reviews!

PR #195, opened 2021-03-08
Rename: unsat-core -> unsat-assumptions

makaimann:
`unsat-assumptions` is more accurate than `unsat-core`, because the smt-switch interface gives the core based on assumptions provided through `check-sat-assuming` and thus the command is actually analogous to `get-unsat-assumptions` rather than `get-unsat-core`.

This is an interface change that might affect users of smt-switch.


ahmed-irfan (approved):
LGTM

PR #198, opened 2021-03-08
Hack sys.modules correctly

cdonovick:
My work around wasn't playing well with my autocomplete but this work.


makaimann (approved):
Looks good to me, thanks for fixing this!

PR #199, opened 2021-03-09
Return canonical objects for Python enums

makaimann:
This PR adds a dictionary to look up the canonical object for both `SortKind`s and `PrimOp`. This ensures that the object returned by `sort.get_sort_kind()` or `op.prim_op` is identical to the one used to create sorts and terms, respectively. This also guarantees that the `hash` works as intended.


cdonovick (approved):
Looks good to me

makaimann:
Thanks for the review! I'll wait for Travis then merge.

PR #201, opened 2021-03-09
Use CMake macro for adding tests

makaimann:
This PR uses a CMake macro to make it less verbose to add a test.


yoni206 (approved):
Looks good, but i am wondering -- should the CMakelists files of the tests for the individual solvers be changed as well?

makaimann:
Yeah that's a good idea. You're saying just make it a uniform style, right?

yoni206:
yes. though not necessary for this PR.

makaimann:
Sounds good, I think I will do that in a separate PR after https://github.com/makaimann/smt-switch/pull/185#pullrequestreview-609026575 is merged which adds a new test directory.

PR #203, opened 2021-03-10
Support set-opt and set-logic in msat backend

makaimann:
This PR adds support for set-opt and set-logic in the msat backend. The MathSAT API requires setting all the options before creating the environment. To accommodate this, the env is lazily initialized when it is needed. Prior to that, options and the logic can be set on the `msat_config` object.

It's worth considering requiring the logic and options up front when creating a solver in the future. The current implementation is more like smt-lib but might make less sense in the API. At least MathSAT and Z3 have the same limitation regarding setting options.

### DIFF
--- a/msat/include/msat_solver.h
+++ b/msat/include/msat_solver.h
@@ -41,13 +41,18 @@ namespace smt {
 class MsatSolver : public AbsSmtSolver
 {
  public:
-  // constructor does basically nothing
-  // but in mathsat factory, MUST setup_env
-  // this is done after constructing because need to call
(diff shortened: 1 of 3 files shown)


ahmed-irfan:
> It's worth considering requiring the logic and options up front when creating a solver in the future. The current implementation is more like smt-lib but might make less sense in the API. At least MathSAT and Z3 have the same limitation regarding setting options.

Another option is to expose the initialize_env() method for each solver.

ahmed-irfan (approved):
LGTM.

PR #207, opened 2021-03-10
Tree walker implementation

akgperson:


README.md
```diff
@@ -1,5 +1,6 @@
 # Smt-Switch
 A generic C++ API for SMT solving. It provides abstract classes which can be implemented by different SMT solvers.
+ADDITION
```
yoni206, line 3:
```suggestion
```

include/term.h
```diff
@@ -170,6 +170,8 @@ using TermVec = std::vector<Term>;
 using TermList = std::list<Term>;
 using UnorderedTermSet = std::unordered_set<Term>;
 using UnorderedTermMap = std::unordered_map<Term, Term>;
+//new type of map for tree walker - TODO put this here or in treewalker.h ??
```
yoni206, line 173: in general, there should be no `TODO`s, so plz `grep` the repo for `TODO`s.

include/tree_walker.h
```diff
@@ -0,0 +1,97 @@
...
+   */
+enum TreeWalkerStepResult
+{
+ TreeWalker_Continue=0,
+ TreeWalker_Skip,
+ TreeWalker_Abort
+};
+
+/** \class
+ * IdentityWalker class.
```
yoni206, line 24: Should be updated. same for all comments in the file.

src/tree_walker.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+//DELETE
+std::string vectorToString(vector<int> v){
+  std::string s;
+  for (int i:v){
+    s += std::to_string(i);
+    s += ", ";
+  }
+  return s;
+}
```
yoni206, line 20: really delete :)

src/tree_walker.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+pair<Term, vector<int>> TreeWalker::visit(Term & node)
+{
+  //cout << "starting visit()" << endl;
```
yoni206, line 24: Generally remove all:
1. commented out 
2. printing outside of tests

src/tree_walker.cpp
```diff
@@ -0,0 +1,195 @@
...
+  int child_no;  //iterates over children, for tracking which child on & path
```
yoni206, line 25: Generally: comments are usually above the line, not in the side.

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +174,511 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+   // EXPECT_EQ(x.second.first, expected_map[x.first].first); //testing equivalence of first element in pair: the topmost node
+    //ASSERT_EQ(x.second.second.size(), expected_map[x.first].second.size());
+  //  for (int i = 0; i < x.second.second.size(); i++){
+    //  EXPECT_EQ(x.second.second[i], expected_map[x.first].second[i]);
+   // }
+    //}
+  }
+}
+
+
```
yoni206, line 681: Please locally: install all solvers but z3:
- msat you already have
- cvc4: using the contrib script
- boolector: using the contrib script
- yices: follow the instrucitons in the contrib script

and then configure with `--msat --cvc4 --yices2 --btor --debug`.

yoni206 (changes requested):
Thanks for this!
First superficial review, after these comments are addressed I will make a deeper pass.

include/term.h
```diff
@@ -170,6 +170,8 @@ using TermVec = std::vector<Term>;
...
+using UnorderedTermPairMap = std::unordered_map<Term, std::pair<Term, std::vector<int>>>;
```
yoni206, line 174: Please move this line to tree_walker.h

include/tree_walker.h
```diff
@@ -0,0 +1,97 @@
...
+ /** Visit a term and all its subterms in a post-order traversal
+  *  the member variable preorder_ is true if it's the first time seeing
+  *  a subterm and false if the traversal is in post-order already
+  *  @param term the term to visit
+  *  @return the term after visiting (returns the value of cache[term]
+  *     -- if it has been cached and returns term otherwise)
+  */
+// smt::Term visit(std::pair <smt::Term, std::vector<int>> & occurrence);
+ std::pair<smt::Term, std::vector<int>> visit(smt::Term & node);
+ using TermPairVec = std::vector<std::pair<Term, int>>;
```
yoni206, line 54: move to the top please, try to follow the structure of other fides.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+  if (query_cache(node, out))
+  {
+    // cache hit
+    return out;
+  }
+
+  //tree_path currently empty, process top node
+  visit_term(node, node, tree_path);
+  TermPairVec to_visit;
+  pair<Term, int> p1 (node, -1);
```
yoni206, line 51: document in detail what does `-1` mean.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+  to_visit.push_back(p1);
+  child_no = 0;
+  for (auto ttt : node){
```
yoni206, line 54:
```suggestion
// add children to `to_visit`.
  for (auto ttt : node){
```

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+  Term current_term;
+  pair<Term, int> pn;
+  while(to_visit.size())
+  {
+    current_pair = to_visit.back();
+    current_term = current_pair.first;
+    child_no = current_pair.second;
+    to_visit.pop_back();
+
+    if (child_no != -1){
```
yoni206, line 73: here explain (again, but in less detail) what does it mean that we see a -1.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+      tree_path.push_back(child_no);
+      visit_term(node, current_term, tree_path);
+      pn.first = current_term;
+      pn.second = -1;
```
yoni206, line 77: explain why are we setting this to -1

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+      to_visit.push_back(pn);
+      child_no = 0;
+      for (auto tt : current_term){
```
yoni206, line 80:
```suggestion
      // adding children to `to_visit`
      for (auto tt : current_term){
```

src/tree_walker.cpp
yoni206, line 54: We have a very similar loop inside the main loop. So please explain (in detail) why do we do this twice, why can't we just add only in the loop.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+        pn.first = tt;
+        pn.second = child_no;
+        to_visit.push_back(pn);
+        child_no++;
+      }
+    }
+    else {
```
yoni206, line 87: plz document what does it mean that we reached the `else` clause.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+      if (!tree_path.empty()){
```
yoni206, line 88: explain why to we pop only non-empty paths.

include/tree_walker.h
```diff
@@ -0,0 +1,97 @@
...
+
+protected:
+ /** Visit a single term.
+  *  Implement this method in a derived class to change the behavior
+  *  of the walker
+  *  @param term the term to visit
+  *  @return a WalkerStepResult to tell the visit method how to proceed
+  */
```
yoni206, line 62: Additionally, please mention that the implementation includes a default function, and also explain what does this default function do (maps every term to its "first observed" occurrence.

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+  Op op = term->get_op();
+  if (!op.is_null())
+  {
+    TermVec cached_children;
+    Term c;
+    pair <Term, vector<int>> occurrence;
+    for (auto t : term)
+    {
+      //should we still query_cache here?
```
yoni206, line 111: I don't know :)

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+    pair <Term, vector<int>> occ;
+    Term occ_key;
+    occ.first = formula;
+    occ_key = solver_->make_term(op, cached_children);
+    occ.second = path;
+    save_in_cache(occ_key, occ);
+  }
+  else
+  {
+      // just keep the leaves the same
```
yoni206, line 123: plz more details

src/tree_walker.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  // finished the traversal
+  // return the cached term if available
+  // otherwise just returns the original term
+  query_cache(node, out);
+  return out;
+}
+
+TreeWalkerStepResult TreeWalker::visit_term(Term & formula, Term & term, vector<int> & path)
```
yoni206, line 101: please add documentation inside this function

tests/unit/unit-walker.cpp
```diff
@@ -21,12 +21,19 @@
...
 using namespace std;
 
 namespace smt_tests {
 
+//class DummyTreeWalker : TreeWalker {
+  //TreeWalkerStepResult visit_term(smt::Term & formula, smt::Term & term, std::vector<int> & path) override{
+    //return TreeWalker_Continue;
+  //}
+//};
+
```
yoni206, line 36:
```suggestion
```

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
 
+class FreshTreeWalker : public TreeWalker {
+  int b_iter {};
+
+  using TreeWalker::TreeWalker;
+
+//  FreshTreeWalker(const smt::SmtSolver & solver,
+  //               bool clear_cache,
+    //             smt::UnorderedTermPairMap * ext_cache = nullptr)
+      //: solver_(solver), clear_cache_(clear_cache), ext_cache_(ext_cache){};
```
yoni206, line 62:
```suggestion
```

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
+    if (!op.is_null())
+    {
+      TermVec cached_children;
+      Term c;
+      pair <Term, vector<int>> occurrence;
+      for (auto t : term)
+      {
+       // occurrence.first = t;
+        //occurrence.second = path;
+        //query_cache(t, occurrence);
```
yoni206, line 82:
```suggestion
```

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
     funsort = s->make_sort(FUNCTION, SortVec{ bvsort, bvsort });
     arrsort = s->make_sort(ARRAY, bvsort, bvsort);
+    boolsort = s->make_sort(BOOL);
   }
   SmtSolver s;
-  Sort bvsort, funsort, arrsort;
+  Sort bvsort, funsort, arrsort, boolsort;
 };
 
+class FreshTreeWalker : public TreeWalker {
```
yoni206, line 54: Above this line, please explain in detail what is the purpose of this walker (something like assigning a fresh Boolean variable to each occurrence of each term).

yoni206, line 54: Maybe change the name to `IndicatorTreeWalker`.

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
+  int b_iter {};
+
+  using TreeWalker::TreeWalker;
+
+//  FreshTreeWalker(const smt::SmtSolver & solver,
+  //               bool clear_cache,
+    //             smt::UnorderedTermPairMap * ext_cache = nullptr)
+      //: solver_(solver), clear_cache_(clear_cache), ext_cache_(ext_cache){};
+
+  TreeWalkerStepResult visit_term(smt::Term & formula, smt::Term & term, std::vector<int> & path) override{
```
yoni206, line 64: Please add detailed documentation to the implemetation of this function.

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
+    Sort boolsort = solver_->make_sort(BOOL);
+
+    string var_name = string("b") + to_string(b_iter);
+    Term b = solver_->make_symbol(var_name, boolsort);
+    b_iter++;
+
+    Op op = term->get_op();
+    if (!op.is_null())
+    {
+      TermVec cached_children;
```
yoni206, line 75: explain what does it mean that we rearched the "if" clause.

tests/unit/unit-walker.cpp
```diff
@@ -38,11 +45,62 @@ class UnitWalkerTests : public ::testing::Test,
...
+      Term c;
+      pair <Term, vector<int>> occurrence;
+      for (auto t : term)
+      {
+       // occurrence.first = t;
+        //occurrence.second = path;
+        //query_cache(t, occurrence);
+        cached_children.push_back(t);
```
yoni206, line 83: these are not cached. Same for the default implementation, so please change the name in both plasces.

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +173,490 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+}
+
+string vectorToString(vector<int> v){
+  string s;
+  for (int i:v){
+    s += to_string(i);
+    s += ", ";
+  }
+  return s;
+}
```
yoni206, line 189: plz delete

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +173,490 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+
+
+void printMap(UnorderedTermPairMap const &m){
```
yoni206, line 192: dito

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +173,490 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+  for (auto const& pair : m){
+    auto val_pair = pair.second;
+    cout << pair.first << ": " << "<" << val_pair.first << ", (" << vectorToString(val_pair.second) << ")>" << " }\n";
+    //cout << "{ " << pair.first << ": " << pair.second << " }" << endl;
+  }
+}
+
+TEST_P(UnitWalkerTests, SimpleTree)
```
yoni206, line 200: In each test:
1. general documentation of what each block of code does.
2. When you create a term, document what term are you creating. For example, before the block of code that creates the term `x+y=x+1`, write a comment like:
```
// construct `x+y=x+1`
```
No need to do this for the sub-terms, though. 

2. When you create an "expected" variable (vector/map/whatever), plz document how it looks like in a comment, e.g.: 
```
x+y-> (x+y, [])
x -> (x+y, [0]
...
```

yoni206 (changes requested):

yoni206 (changes requested):

src/tree_walker.cpp
```diff
@@ -0,0 +1,197 @@
...
+  child_no = 0;
+  //push_back all of topmost node's children to prepare for the loop
+  for (auto ttt : node){
+    p1.first = ttt;
+    p1.second = child_no;
+    to_visit.push_back(p1);
+    child_no++;
+  }
+
+//  Term t;
```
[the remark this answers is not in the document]
akgperson: I removed the line with res... because it doesn't look like it is actually getting used at all... it must have been removed at some point...

src/tree_walker.cpp
akgperson: Is this necessary or is the line above the start of the for loop sufficient? Should that line be moved to within the for loop?

yoni206: I don't see any line above the loop currently, but anyway it is enough to have it a single time above the loop.

tests/unit/unit-walker.cpp
```diff
@@ -37,12 +38,57 @@ class UnitWalkerTests : public ::testing::Test,
...
   SmtSolver s;
-  Sort bvsort, funsort, arrsort;
+  Sort bvsort, funsort, fun2sort, arrsort, boolsort;
 };
 
+/* Custom Tree Walker that builds up a map assigning fresh boolean variables to each occurrence of each term in the formula it visits.
+ * One example of overriding visit_term to customize the behavior of the walker. */
+class IndicatorTreeWalker : public TreeWalker {
+  //iterator with which to name fresh variables: b0, ..., bn
+  int b_iter {};
```
yoni206, line 53:
```suggestion
  int b_iter;
```

tests/unit/unit-walker.cpp
```diff
@@ -37,12 +38,57 @@ class UnitWalkerTests : public ::testing::Test,
...
   }
   SmtSolver s;
-  Sort bvsort, funsort, arrsort;
+  Sort bvsort, funsort, fun2sort, arrsort, boolsort;
 };
 
+/* Custom Tree Walker that builds up a map assigning fresh boolean variables to each occurrence of each term in the formula it visits.
+ * One example of overriding visit_term to customize the behavior of the walker. */
+class IndicatorTreeWalker : public TreeWalker {
+  //iterator with which to name fresh variables: b0, ..., bn
```
yoni206, line 52:
```suggestion
  //counter with which to name fresh variables: b0, ..., bn
```

tests/unit/unit-walker.cpp
```diff
@@ -37,12 +38,57 @@ class UnitWalkerTests : public ::testing::Test,
...
+    boolsort = s->make_sort(BOOL);
   }
   SmtSolver s;
-  Sort bvsort, funsort, arrsort;
+  Sort bvsort, funsort, fun2sort, arrsort, boolsort;
 };
 
+/* Custom Tree Walker that builds up a map assigning fresh boolean variables to each occurrence of each term in the formula it visits.
+ * One example of overriding visit_term to customize the behavior of the walker. */
+class IndicatorTreeWalker : public TreeWalker {
```
yoni206, line 51: plz specify in which test this class will be used.

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +161,730 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+  // traversal of simple formula using TreeWalker, test that cache gets built up as expected
+
+  SolverConfiguration sc = GetParam();
+  if (sc.is_logging_solver) {
+    // test only valid for logging solver
+
+    Term x = s->make_symbol("x", bvsort);
+    Term xp1 = s->make_term(BVAdd, x, s->make_term(1, bvsort));
+
+    UnorderedTermPairMap UndMap;
```
yoni206, line 175: rename to `und_map` or any other name, but with lower-case-underscores-style. Same for all other occurrences in all other tests of this variable.

tests/unit/unit-walker.cpp
```diff
@@ -115,6 +161,730 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+    // test equivalence of expected_map and cache populated during the traversal
+    for (auto const& x : UndMap){
+      // testing equivalence of topmost node (first element in pair)
+      EXPECT_EQ(x.second.first, expected_map[x.first].first);
+      ASSERT_EQ(x.second.second.size(), expected_map[x.first].second.size());
+      // testing path equivalence
+      for (int i = 0; i < x.second.second.size(); i++){
+        EXPECT_EQ(x.second.second[i], expected_map[x.first].second[i]);
+      }
+    }
```
yoni206, line 213: plz make a helper function for this code, and call it in every test, instead of inlining it.

tests/unit/unit-walker.cpp
akgperson: .

tests/unit/unit-walker.cpp
akgperson: double check on placement w/in file and my instantiation

tests/unit/unit-walker.cpp
akgperson: test fails with this change

tests/unit/unit-walker.cpp
```diff
@@ -37,10 +39,67 @@ class UnitWalkerTests : public ::testing::Test,
...
   }
   SmtSolver s;
-  Sort bvsort, funsort, arrsort;
+  Sort bvsort, funsort, fun2sort, arrsort, boolsort;
+};
+
+/* Custom Tree Walker that builds up a map assigning fresh boolean variables to
+ * each occurrence of each term in the formula it visits. One example of
+ * overriding visit_term to customize the behavior of the walker.
+ * Class gets used in last 9 tests: SimpleTree, PathDecomp, PathTests1,
```
yoni206, line 53: i only see this class used in one test, not 9.

tests/unit/unit-walker.cpp
```diff
@@ -37,10 +39,67 @@ class UnitWalkerTests : public ::testing::Test,
...
+   * term in the formula, where the occurrences are represented as a  pair
+   * containing the formula the term appears in and the path indicating the
+   * term's location in the formula it appears in. For example, the formula
+   * x+2=y traversed with this custom version of visit_term will build the
+   * corresponding map containing: b0 -> <x+2=y, []>, corresponding to the term
+   * x+2=y b1 -> <x+2=y, [1]>, corresponding to the term y b2 -> <x+2=y, [0]>,
+   * corresponding to the term x+2 b3 -> <x+2=y, [0,1]>, corresponding to the
+   * term 2 b4 -> <x+2=y, [0,0]>, corresponding to the term x As in this
+   * example, each occurrence of each term in the formula should correspond to a
+   * fresh variable mapping to its occurrence in the formula
```
yoni206, line 74: i guess clang-format deleted some new-lines here. Please re-format this part yourself, and in the next commits we will use git-clang-format instead of clang-format, so that only new changes are re-formatted.

tests/unit/unit-walker.cpp
```diff
@@ -115,8 +174,743 @@ TEST_P(UnitWalkerTests, FunSubstitution)
   EXPECT_EQ(fy, iw.visit(fx));
 }
 
-INSTANTIATE_TEST_SUITE_P(ParametrizedUnitWalker,
-                         UnitWalkerTests,
-                         testing::ValuesIn(filter_solver_configurations({ TERMITER })));
+/* helper function to test equivalency of passed_map that TreeWalker builds up
+ * against expected_map that should have been built up gets used for all tests
```
yoni206, line 178:
```suggestion
 * against expected_map that should have been built up. gets used for all tests
```

include/tree_walker.h
```diff
@@ -0,0 +1,134 @@
+#pragma once
```
makaimann, line 1: Could you add a file header above this line. You can just copy-paste the block comment from here: https://github.com/makaimann/smt-switch/blob/68a0a4180aefb4de756178c7a42b79d73db1243e/include/identity_walker.h#L2. Then replace the filename, \brief description, and put your name there.

include/tree_walker.h
```diff
@@ -0,0 +1,134 @@
...
+namespace smt {
+/* vector of pairs holding terms and ints that gets used within visit in the
+ * vector to_visit as a means of storing what terms we need to visit along with
+ * each term's child number relative to its parent node or a "-1 flag" that
+ * indicates all of a node's children have already been visited.
+ */
+using TermPairVec = std::vector<std::pair<Term, int>>;
+/* used to map from terms to their occurrences, represented as a pair of the
+ * formula it is found in and the path indicating the terms placement in the
+ * formula it is find in. A term's path is built up as list of indices where
```
makaimann, line 17: minor: find -> found

include/tree_walker.h
```diff
@@ -0,0 +1,134 @@
...
+ */
+using UnorderedTermPairMap =
+    std::unordered_map<Term, std::pair<Term, std::vector<int>>>;
+
+/** \enum
+ * Walker_Continue : rebuild the current term and continue
+ * Walker_Skip     : skip this term and all subterms
+ * Walker_Abort    : abort visiting
+ */
+enum TreeWalkerStepResult
```
makaimann, line 31: This is not necessary, but I think it might be nice to use the same enums since they're identical. One way to do this is to create a new header file `walkers.h` or something like that and put the enum there and include that file in both `identity_walker.h` and `tree_walker.h`.

include/tree_walker.h
```diff
@@ -0,0 +1,134 @@
...
+   *     1: [0,1]
+   *     2: [1]
+   *  @param formula the term taken to visit in visit, the topmost node of the
+   * visited formula
+   *  @param term the term to visit in visit_term, a particular sub-term of the
+   * formula (some node in the tree representation of the full formula being
+   * traversed in visit)
+   *  @param path the path for the particular term in formula, which we are
+   * visiting
+   *  @return a WalkerStepResult to tell the visit method how to proceed
```
makaimann, line 96: This should be `TreeWalkerStepResult` in the comment, unless you unify `TreeWalkerStepResult` and `WalkerStepResult`.

tests/unit/unit-walker.cpp
```diff
@@ -115,8 +174,743 @@ TEST_P(UnitWalkerTests, FunSubstitution)
...
+    Term fullform = s->make_term(Equal, xpx, rhs);
+
+    UnorderedTermPairMap passed_map;
+    IndicatorTreeWalker itw(s, false, &passed_map);
+
+    pair<Term, vector<int>> p1;
+    p1 = itw.visit(fullform);
+
+    /* expected map:
+     * b0: <(x+x)={(y+y)+[(x+2)-(x+x)]} , []>, where b0 corresponds to
```
makaimann, line 735: I don't totally follow the format of this comment. Are the keys all the `b*`? I think the `..., where b* corresponds to` is confusing me. It's also possible `clang-format` messed this up (that's happened to me before).

makaimann (approved):
Looks really good to me, thanks for working on this! It's well documented and looks like the test is pretty extensive. Left some minor comments but none of them are critical.

include/tree_walker.h
makaimann: Talked on slack with Yoni and decided this should be delayed to a separate PR, so no action necessary. Resolving this conversation.

tests/unit/unit-walker.cpp
makaimann: Okay this was explained to me offline, makes sense! Only small thing might be to start a new line for each key, value pair. E.g.:
```
* b0: <(x+x)={(y+y)+[(x+2)-(x+x)]} , []>, where b0 corresponds to
* (x+x)={(y+y)+[(x+2)-(x+x)]} 
* b1: <(x+x)={(y+y)+[(x+2)-(x+x)]} , [1]>, where b1 corresponds to 
* {(y+y)+[(x+2)-(x+x)]} 
```

or something like that.

yoni206 (approved):

PR #208, opened 2021-03-11
Check sat assuming msat

makaimann:
This PR would support arbitrary formulas in check-sat-assuming in the mathsat backend. The motivation for this is that most other solvers allow this and there's a good argument that it is consistent with the SMT-LIB standard. SMT-LIB requires that the arguments to check-sat-assuming be symbols. But those symbols could be from a `define-fun` which might make that symbol an alias for an arbitrary term. Thus in the text interface, it is basically just requiring that the assumption has a _name_. Through the API we are using the objects directly and thus it makes sense to be able to use any arbitrary term.

This PR simulates the more general interface by creating labels as needed. To avoid polluting the global assertions with lots of labels, it periodically resets the environment (if at context 0). This also requires keeping track of "actual" assertions and re-adding them after resetting. I set the number of clauses to consider worth clearing to 10,000.

### DIFF
--- a/msat/include/msat_solver.h
+++ b/msat/include/msat_solver.h
@@ -137,6 +137,11 @@ class MsatSolver : public AbsSmtSolver
   // for interacting with third-party MathSAT-specific software
   msat_env get_msat_env() const { return env; };
 
+  // getters and setters for advanced use / testing
+  size_t max_assump_clauses() const { return max_assump_clauses_; }
+
+  void set_max_assump_clauses(size_t m) { max_assump_clauses_ = m; }
+
  protected:
   msat_config cfg;
   // marked mutable because want to stick with const interface for functions
--- a/msat/include/msat_solver.h
+++ b/msat/include/msat_solver.h
@@ -204,6 +204,7 @@ class MsatSolver : public AbsSmtSolver
       // check that label is cached correctly
       assert(msat_term_id(lbl) == msat_term_id(label(ma)));
       msat_assert_formula(env, msat_make_or(env, msat_make_not(env, lbl), ma));
+      num_assump_clauses_++;
       assumption_map_[msat_term_id(lbl)] = ma;
       lbls.push_back(lbl);
     }
--- a/msat/include/msat_solver.h
+++ b/msat/include/msat_solver.h
@@ -58,7 +58,9 @@ class MsatSolver : public AbsSmtSolver
         valid_model(false),
         logic(""),
         num_assump_clauses_(0),
-        max_assump_clauses_(10000){};
+        max_assump_clauses_(10000)
+    {
+    };
   MsatSolver(const MsatSolver &) = delete;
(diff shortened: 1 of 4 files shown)


msat/include/msat_solver.h
```diff
@@ -42,17 +42,21 @@ class MsatSolver : public AbsSmtSolver
...
-    : AbsSmtSolver(MSAT),
-      cfg(msat_create_config()),
-      env_uninitialized(true),
-      logic(""){};
+      : AbsSmtSolver(MSAT),
+        cfg(msat_create_config()),
+        env_uninitialized(true),
+        logic(""),
+        num_assump_clauses_(0),
+        max_assump_clauses_(10000){};
```
ahmed-irfan, line 50:
```suggestion
        max_assump_clauses_(10000)
    {
    };
```

msat/include/msat_solver.h
```diff
@@ -42,17 +42,21 @@ class MsatSolver : public AbsSmtSolver
...
   MsatSolver(msat_config c, msat_env e)
       : AbsSmtSolver(MSAT),
         cfg(c),
         env(e),
         env_uninitialized(false),
         valid_model(false),
-        logic(""){};
+        logic(""),
+        num_assump_clauses_(0),
+        max_assump_clauses_(10000){};
```
ahmed-irfan, line 59:
```suggestion
        max_assump_clauses_(10000)
    {
    };
```

msat/src/msat_solver.cpp
```diff
@@ -205,31 +212,16 @@ Result MsatSolver::check_sat()
 Result MsatSolver::check_sat_assuming(const TermVec & assumptions)
```
ahmed-irfan, line 212: isnt it duplicate? we have an inline function for this in the header file. Or am I missing something?

ahmed-irfan: oh, i see now, the argument is different

ahmed-irfan: how can we make it more clear?

ahmed-irfan (changes requested):
Left few comments and minor suggestions

msat/src/msat_solver.cpp
makaimann: Yeah good point, we could give it a different name? Like -- `msat_check_sat_assuming` or something?

ahmed-irfan: maybe give the other one the name : check_sat_assuming_msatvec?

makaimann: Cool, done with 948f7dcd7c89f478ff44fb847d7c8db7b1b430d5

makaimann:
Link to Travis run: https://travis-ci.org/github/makaimann/smt-switch/builds/763128818. It's still not syncing for some reason, I'll look into that soon.

ahmed-irfan (approved):

PR #213, opened 2021-03-12
Bug fixes from testing suite

lstuntz:
Various new implementations and bug fixes as uncovered from the z3 testing suite. There are 4 tests that don't pass in this pr, fixes for those will be in a later pr.


lstuntz:
@makaimann There is one bug that is in this pr that is worth noting if you are going to be doing additional testing/experiments. The is_value function in z3_terms.cpp is wonky -- I'm working on fixing it but it's not fixed in this pr

z3/src/z3_solver.cpp
```diff
@@ -85,12 +86,74 @@ const unordered_map<PrimOp, variadic_fun> z3_variadic_ops({
...
+    msg += " - not implemented for Z3 backend.";
+    throw NotImplementedException(msg.c_str());
+  }
 }
 
 void Z3Solver::set_logic(const std::string logic)
 {
-  throw NotImplementedException("Set logic not implemented for Z3 backend.");
+  const char * l = logic.c_str();
+  slv = solver(ctx, l);
```
makaimann, line 156: So this is Z3-specific question. Do you need to do anything to clean up the old solver?

I think this is good, but in case it's relevant I implemented `set_logic` in mathsat which has a similar API with https://github.com/makaimann/smt-switch/pull/203. I could have reconstructed it like you're doing here, but it seems like mathsat leaks memory if you reconstruct, so instead I waited to construct the env (equivalent of the solver) until after all options / logic is set. A little heavy-handed, but technically works.

z3/src/z3_solver.cpp
```diff
@@ -319,13 +382,26 @@ Result Z3Solver::check_sat_assuming_set(const UnorderedTermSet & assumptions)
...
 Term Z3Solver::get_value(const Term & t) const
 {
-  throw NotImplementedException("Get value not implemented for Z3 backend.");
+  shared_ptr<Z3Term> zterm = static_pointer_cast<Z3Term>(t);
+  if (zterm->is_function)
+  {
+    throw IncorrectUsageException("Cannot evaluate a function.");
+  }
+  z3::model model = slv.get_model();
+  expr eval = model.eval(zterm->term, true);
```
makaimann, line 403: Just for my own understanding, what is the second parameter here? I saw that it's `model_completion`, but what does that mean?

z3/src/z3_term.cpp
```diff
@@ -104,26 +104,159 @@ bool Z3Term::compare(const Term & absterm) const
 
 Op Z3Term::get_op() const
 {
-  throw NotImplementedException("get_op not yet implemented.");
+  if (is_function)
+  {
+    throw IncorrectUsageException("A function has no operator.");
```
makaimann, line 109: Here you can just return a null operator: `return Op()`. In smt-switch, you can still get the operator, it's just null.

z3/src/z3_term.cpp
```diff
@@ -104,26 +104,159 @@ bool Z3Term::compare(const Term & absterm) const
...
+  }
+  else
+  {
+    if (!term.is_app())
+    {
+      throw IncorrectUsageException(
```
makaimann, line 115: This should also return a null operator I think (`get_op` shouldn't throw exceptions). What kind of terms have false for `is_app()`?

z3/src/z3_term.cpp
```diff
@@ -104,26 +104,159 @@ bool Z3Term::compare(const Term & absterm) const
...
+        case Z3_OP_ZERO_EXT: return Op(Zero_Extend);
+        case Z3_OP_SIGN_EXT: return Op(Sign_Extend);
+        case Z3_OP_REPEAT: return Op(Repeat);
+        case Z3_OP_INT2BV: {
+          size_t out_width = range.bv_size();
+          return Op(Int_To_BV, out_width);
+        }
+        case Z3_OP_BV2INT:
+          return Op(BV_To_Nat);
+          // Op(Apply) not handled...
```
makaimann, line 199: If there's a way to check whether something is a function application, you could do that above. Doesn't have to be in this PR, just mentioning it.

z3/src/z3_term.cpp
```diff
@@ -104,26 +104,159 @@ bool Z3Term::compare(const Term & absterm) const
...
 
 bool Z3Term::is_symbolic_const() const
 {
-  return (term.is_const() && !is_function);
+  if (is_function || is_param())
+  {
+    return false;
+  }
+  return is_symbol();
+  // return (!is_function && term.is_const());
```
makaimann, line 245: minor: old comment can be removed

z3/src/z3_term.cpp
```diff
@@ -104,26 +104,159 @@ bool Z3Term::compare(const Term & absterm) const
...
 }
 
 bool Z3Term::is_value() const
 {
-  throw NotImplementedException("is_value not implemented for Z3 backend.");
+  // return (!is_function && term.is_const());
```
makaimann, line 250: minor: old comment can be removed

makaimann (approved):
Looks really good! Thanks for the PR. I just left a few comments: some minor and some questions. The only thing that's actually important would be a small tweak in `get_op`.

PR #216, opened 2021-03-16
Use a pointer to a vector for lists of s-exprs in parser

makaimann:
This PR would use a pointer to a vector for representing a list of sexprs. This is to avoid copying the vector. For some reason, marking the semantic value as a reference does not seem to work.


makaimann:
Sorry for the force pushes, I didn't realize I wasn't branched off of master.

ahmed-irfan (approved):
LGTM

makaimann:
Thanks for the review!

PR #219, opened 2021-03-17
Add initial support for Bitwuzla backend

makaimann:
This PR adds basic support for a [Bitwuzla](https://bitwuzla.github.io) backend.


include/solver.h
```diff
@@ -352,6 +352,9 @@ class AbsSmtSolver
   virtual Term substitute(const Term term,
                           const UnorderedTermMap & substitution_map) const;
 
+  virtual TermVec substitute_terms(
```
yoni206, line 355: What's the motivation to add this to `AbsSolver`? Do most other solvers support this natively? Is this expected to be commonly used? If there isn't a good motivation, I wouldn't include this.

include/solver_enums.h
```diff
@@ -63,6 +63,8 @@ enum SolverAttribute
   THEORY_DATATYPE,
   // supports quantifiers
   QUANTIFIERS,
+  // supports getting the operator of a term
+  GET_OP,
```
yoni206, line 67: who doesn't support this?

python/smt_switch/enums_dec.pxi
```diff
@@ -40,7 +40,6 @@ cdef extern from "solver_enums.h" namespace "smt":
     cdef cppclass c_SolverAttribute "smt::SolverAttribute":
         pass
 
-    cdef c_SolverAttribute c_LOGGING "smt::LOGGING"
```
yoni206, line 43: my bad for not removing this a while ago, sorry :)

python/smt_switch/enums_imp.pxi
```diff
@@ -151,10 +151,6 @@ solverattr = ModuleType('solverattr')
 _add_module(solverattr)
 solverattr.__file__ = FILENAME
 
-cdef SolverAttribute LOGGING = SolverAttribute()
-LOGGING.sa = c_LOGGING
-setattr(solverattr, "LOGGING", LOGGING)
-
 cdef SolverAttribute TERMITER = SolverAttribute()
 TERMITER.sa = c_TERMITER
 setattr(solverattr, "TERMITER", TERMITER)
```
yoni206, line 156: Assuming this new `get_op` attribute is indeed added to c++, shouldn't it be added to python?

yoni206 (approved):
Looks good! Left some comments for your consideration. Nothing is critical, so feel free to merge. Additionally -- it would be nice to have a test file that is specific for bitwuzla (like all other solvers have).

include/solver_enums.h
makaimann: Good point, it used to be CVC4 but that has changed. And @lstuntz just added support for it in the Z3 backend so currently all solvers support it. Addressed with 2e5a182f6851b357b71f8bd9380efc077cfae34e.

CMakeLists.txt
```diff
@@ -121,6 +121,18 @@ if (BUILD_BTOR)
   add_definitions(-DBTOR_HOME=${BTOR_HOME})
 endif (BUILD_BTOR)
 
+if (BUILD_BITWUZLA)
```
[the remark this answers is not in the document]
makaimann: Thanks for catching that! Added with 83ced930ccdbf06c270a45c09406653fa4e7f60b.

bitwuzla/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+add_library(smt-switch-bitwuzla "${SMT_SWITCH_LIB_TYPE}"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/bitwuzla_factory.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/bitwuzla_solver.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/bitwuzla_sort.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/bitwuzla_term.cpp"
+  # "${PROJECT_SOURCE_DIR}/contrib/memstream-0.1/memstream.c"
```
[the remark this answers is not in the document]
makaimann: Yep, done with eaaa75a3447552909c0f5ad27a3ca5090dd23dc9.

bitwuzla/include/bitwuzla_solver.h
```diff
@@ -0,0 +1,154 @@
...
+                  const UnorderedTermMap & substitution_map) const override;
+  TermVec substitute_terms(
+      const TermVec & term,
+      const UnorderedTermMap & substitution_map) const override;
+  void dump_smt2(std::string filename) const override;
+
+  // getters for solver-specific objects
+  // for interacting with third-party Bitwuzla-specific software
+
+  Bitwuzla * get_bitwuzla() const { return bzla; };
```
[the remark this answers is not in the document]
makaimann: Yeah it should, that's for expert users. Each solver can optionally expose what it wraps. This lets us do fancy things, for example get the CVC4 solver to use the SyGuS features.

include/solver.h
makaimann: There's actually a pretty good case for this even if the solver doesn't support it. If you're using the same substitution map for replacements in several terms, it's expensive to copy and unpack it each time (for model checking we could have thousands of elements in the map). If the solver backend _does_ support it, then you also get the benefit of a shared cache.

python/smt_switch/enums_imp.pxi
makaimann: Yep, but I decided to remove it based on your feedback so we're good!

makaimann:
Thanks for the review @yoni206! I've addressed all your comments.

bitwuzla/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  )
+
+target_include_directories (smt-switch-bitwuzla PUBLIC "${PROJECT_SOURCE_DIR}/include")
+target_include_directories (smt-switch-bitwuzla PUBLIC "${PROJECT_SOURCE_DIR}/contrib/memstream-0.1/")
+target_include_directories (smt-switch-bitwuzla PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include")
+target_include_directories (smt-switch-bitwuzla PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/bitwuzla/include")
+target_include_directories (smt-switch-bitwuzla PUBLIC "${BITWUZLA_HOME}/src/api/c")
+
+target_link_libraries(smt-switch-bitwuzla "${BITWUZLA_HOME}/build/lib/libcbitwuzla.a")
+target_link_libraries(smt-switch-bitwuzla "${BITWUZLA_HOME}/deps/cadical/build/libcadical.a")
```
aniemetz, line 17: Just FYI, for skeleton preprocessing, Bitwuzla still needs Lingeling (as does Boolector). We plan to switch this to CaDiCaL, but right now it isn't yet. If Bitwuzla is not built with Lingeling, skeleton preprocessing will not be performed.

bitwuzla/src/bitwuzla_solver.cpp
```diff
@@ -0,0 +1,694 @@
...
+  if (sk == ARRAY)
+  {
+    return make_shared<BzlaSort>(
+        bitwuzla_mk_array_sort(bzla, bsort1->sort, bsort2->sort));
+  }
+  else if (sk == FUNCTION)
+  {
+    vector<BitwuzlaSort *> domain_sorts({ bsort1->sort });
+    return make_shared<BzlaSort>(bitwuzla_mk_fun_sort(
+        bzla, domain_sorts.size(), &domain_sorts[0], bsort2->sort));
```
aniemetz, line 237: Nicer: 
```suggestion
        bzla, domain_sorts.size(), domain.data(), bsort2->sort));
```

bitwuzla/CMakeLists.txt
makaimann: Thanks! What is skeleton preprocessing? It's probably okay to go without it until it's moved to CaDiCaL right?

bitwuzla/src/bitwuzla_solver.cpp
makaimann: Right, thanks! P.s. I have actually been using that since you told me about it and it is much nicer -- I never knew! This was just old.

bitwuzla/CMakeLists.txt
aniemetz: It's pretty much what the non-clausal simplification preprocessing pass does in CVC4. It's a preprocessing technique to derive simplifications on the Boolean layer.

aniemetz: Performance should be better, in general, with skeleton preprocessing.

PR #220, opened 2021-03-17
Various small fixes for Z3 backend

makaimann:
This includes:
* fixes for `make_term(string, sort)`
* fix for `term->to_int`
* adds `BVComp` operator
* compiles Z3 with `--single-threaded` which helped with static linking, but is not a great long term solution
* fixes for quantifiers


contrib/setup-z3.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# issues creating a static binary with z3
+# (related to pthread)
+# see https://github.com/Z3Prover/z3/issues/4554
```
yoni206, line 5: I am not sure what does this comment supposed to mean for the reader.

tests/z3/z3_test.cpp
```diff
@@ -68,7 +68,7 @@ int main()
   cout << concat << endl;
 
   Sort bvsort2 = s->make_sort(BV, 8);
-  Term bvterm2 = s->make_term("x", bvsort2);
+  Term bvterm2 = s->make_symbol("x", bvsort2);
```
yoni206, line 71: maybe make sure this change isn't needed in other parts of the z3-specific tests?

z3/src/z3_solver.cpp
```diff
@@ -7,6 +7,9 @@
 
 #include <iostream>
 
+// TEMP for conversions
```
yoni206, line 10: why is this temp and what will it be replaced after that? I would explain in the comment maybe.

yoni206 (approved):
LGTM with some minor comments, though @lstuntz might have more comments.

contrib/setup-z3.sh
makaimann: Ahh yeah was meant for me, not for other people but that isn't very useful. Fixed with e72a6cb5875bb1f9c4248417ec37bafacaee449e

tests/z3/z3_test.cpp
makaimann: They run so this should be the only place.

z3/src/z3_solver.cpp
makaimann: Good idea, addressed with 6ead7ccfadad9877af9ff457dfab4dd6c4676fa9.

makaimann:
Thanks for the review! I'm actually going to merge this now, because I'd like these updates in master for a new run of experiments. But if you get a chance @lstuntz, you could check this over and create new issues / PRs as needed.

PR #222, opened 2021-03-18
Add an example and instructions / script for running it

makaimann:
This PR adds an example usage of the API with a script and examples for how to run it.

This could have been incorporated into the CMake infrastructure, but I thought it was nice to show how someone might use it outside of CMake for their own project. It gives a better understanding of what happens under the hood.


examples/Makefile
```diff
@@ -0,0 +1,13 @@
+all: cvc4_qf_ufbv btor_qf_ufbv
+
+# Note: assumes smt-switch has been installed in a directory called
+# example-install in this directory
```
yoni206, line 4: maybe mention here that the script `build.sh` does just that?

examples/Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+cvc4_qf_ufbv: cvc4_qf_ufbv.cpp
```
yoni206, line 6: CVC4 will take time to install. This is unrelated to this PR, btu maybe in the artifact you can have it already installed in deps.

examples/README.md
```diff
@@ -0,0 +1,6 @@
+# Examples
+
+This directory contains example usage of smt-switch.
+
+## QF_UFBV
+`QF_UFBV` stands for quantifier-free uninterpreted functions and bit-vectors. The files `cvc4_qf_ufbv.cpp` and `btor_qf_ufbv.cpp` demonstrate some common usage of the API with this logic. The `build.sh` script will setup and install smt-switch in this directory and then build the examples. The script has comments explaining each step for building smt-switch. This assumes that you have all the necessary dependencies other than the underlying solvers and smt-switch already installed.  You can look at the `Makefile` to understand how to link smt-switch to a binary. If the script succeeds in building everything, you can run each of the files with `./cvc4_qf_ufbv.out` and `./btor_qf_ufbv.out`, respectively. Note that there may be a delay for looking up the shared libraries on some systems the first time they are run. Importantly, notice that the files differ by only two lines, demonstrating that changing the solver is straightforward.
```
yoni206, line 6: Can you maybe split this to several lines in a meaningful way?

yoni206, line 6: Maybe write a warning that running the build script will take time, depending on the environment, and give an estimate?

yoni206, line 6: I actually used the script and it failed because i had an old version of CVC4 lying around in deps. To prevent that, maybe install the solvers inside the newly created examples dir?

examples/README.md
```diff
@@ -0,0 +1,6 @@
+# Examples
+
+This directory contains example usage of smt-switch.
+
+## QF_UFBV
```
yoni206, line 5: can u explain here how one can change the cpp recompile and rerun?

yoni206 (commented):
Thanks for doing this, i think it will be very useful for users.

I have some comments inline, and also another one -- maybe link to this file from the main README?

examples/Makefile
makaimann: Good idea -- addressed with c2b862a13b2101020b66ded1942118f9002f718f

examples/Makefile
makaimann: good point, I'll make sure it's installed.

examples/README.md
makaimann: Good idea, done with c2b862a13b2101020b66ded1942118f9002f718f

makaimann: Since we decided offline to separate building the backend solvers this shouldn't be a problem anymore.

makaimann: Ditto. Handled by removing that from the build script.

examples/README.md
makaimann: Great idea -- done with c2b862a13b2101020b66ded1942118f9002f718f.

yoni206 (approved):

PR #225, opened 2021-03-19
Handle Yices2 `Distinct` operator variadicly

Bartvhelvert:
Currently in the Yices2 backend the `Distinct` operation is handled as a pairwise binary comparison even though it can be processed as a variadic operator in the Yices2 backend.


makaimann (approved):
Looks good to me, thanks for the PR!

PR #229, opened 2021-04-07
Switch CI to GitHub Actions

makaimann:


makaimann:
Marked `do not merge` because should turn off Travis hooks and delete the `.travis.yml` file before merging.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        name:
+          [
+            btor,
+            cvc4,
+            msat,
+            yices2,
+            z3
```
yoni206, line 15: no bitwuzla?

.travis.yml
```diff
@@ -75,20 +75,22 @@ jobs:
       env: SOLVER='z3' OPTS=""
 
 script:
+  # disabling Travis for now while evaluating GitHub Actions
```
yoni206, line 78: maybe consider just removing the file? if we need it later we know where to find it

yoni206 (approved):
LGTM modulo 2 optional suggestions.

.github/workflows/ci.yml
makaimann: It's not public yet so there's no good way for me to pull it.

.travis.yml
makaimann: Yeah I'll definitely delete it before merging, thanks!

PR #238, opened 2021-05-17
Add SortingNetwork implementation and test

makaimann:
This PR adds a symbolic boolean sorting network implementation to Smt-Switch


makaimann:
Yep, thanks for catching that! Fixed with cd5ddfd164e898ae9f7016a3da799e4a9dc72c5d

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+#include "smt.h"
+
+namespace smt {
+
+/** \class SortingNetwork
+ *         Implements a symbolic sorting network for boolean-sorted variables.
+ *         All methods are const so it can be re-used for terms from the same
+ *         solver.
+ *         Does not return the actual terms, but is used to determine how many
+ *         variables are assigned to true symbolically.
```
yoni206, line 27: But it does return a list of terms... Can you clarify the comment (in the comment)?

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+ *         Definition:
+ *            A sorting network takes a list of boolean terms
+ *             and returns a list of the same length.
+ *            Let the input be x := [x0, ..., xk],
+ *             then the output of a sorting network is a list
+ *             of boolean terms, y := [y0, ..., yk] such that
+ *             any model assigns yi to true iff i or more
+ *             elements of x were set to true
```
yoni206, line 35:
```suggestion
 *             elements of x were set to true in that model.
```

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+ *            Thus, ignoring the edge cases, Nt = n iff
```
yoni206, line 36: Please consider not ignoring the edge cases :-) Maybe adding them after this explanation. Not critical though.

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+ *            if the input is [x0, x1]
+ *            the output is [Or(x0, x1), And(x0, x1)]
+ *            so that the first term evaluates to true iff one or more
+ *            of the inputs evaluates to true, and the second evaluates
+ *            to true iff both evaluate to true
+ *         Example 2:
+ *            if the input is [x0, x1, x2, x3]
+ *            then the model values for each element in the sorting network
+ *            output for any model that sets 3 of them to true and 1 to false
+ *            would be [true, true, true, false]
```
yoni206, line 49:
```suggestion
 *            would be [true, true, true, false]. Note that these are the model values, not the terms that will be returned from the function.
```

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+   *  @return the (symbolic) sorting network output
+   */
+  TermVec sorting_network(const TermVec & unsorted) const;
+
+  /** Sorts vectors recursively
+   *  Used as helper function for sorting_network
+   *  @param unsorted a vector of boolean terms
+   *  @param symbolically sorted output
+   */
+  TermVec sorting_network_rec(const TermVec & unsorted) const;
```
yoni206, line 68: Should this function be private?

yoni206, line 68: Did you consider implementing a non-recursive function?

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+
+  /** Return symbolic sorting for two terms
+   *  @param t1 the first term
+   *  @param t2 the second term
+   *  @return a length two vector with the symbolic sorting result
+   */
+  TermVec sort_two(const Term & t1, const Term & t2) const;
+
+  /** Merges two symbolically sorted vectors
```
yoni206, line 77: Can you specify what does merging mean?

include/sorting_network.h
```diff
@@ -0,0 +1,88 @@
...
+   *  @param sorted1
+   *  @param sorted2
+   *  @return the combined vector
+   */
+  TermVec merge(const TermVec & sorted1, const TermVec & sorted2) const;
```
yoni206, line 82: Also, make private (or protected)?

src/sorting_network.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  return sorting_network_rec(unsorted);
+}
+
+TermVec SortingNetwork::sorting_network_rec(const TermVec & unsorted) const
+{
+  size_t len = unsorted.size();
+  if (len == 1)
```
yoni206, line 56:
```suggestion
  if (len == 0 || len == 1)
```
Because I fear we might get an infinite loop if `len=0`? You check this in `sorting_network` but what if you keep dividing by 2 and then reach `0`?

src/sorting_network.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return sort_two(sorted1[0], sorted2[0]);
+  }
+
+  bool sorted1_is_even = (len1 % 2) == 0;
+  bool sorted2_is_even = (len2 % 2) == 0;
+
+  if (!sorted1_is_even and sorted2_is_even)
+  {
+    // normalize so we don't have to do all cases
+    return merge(sorted2, sorted1);
```
yoni206, line 106: Did you consider a non-recursive version? Maybe adding an issue for this in the repo?

src/sorting_network.cpp
```diff
@@ -0,0 +1,199 @@
...
+  {
+    if (i % 2 == 0)
+    {
+      odd_sorted2.push_back(sorted2[i]);
+    }
+    else
+    {
+      even_sorted2.push_back(sorted2[i]);
+    }
+  }
```
yoni206, line 137: Can you document what this block is doing?

src/sorting_network.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  TermVec res_even = merge(even_sorted1, even_sorted2);
+  TermVec res_odd = merge(odd_sorted1, odd_sorted2);
+  size_t len_res_even = res_even.size();
+  size_t len_res_odd = res_odd.size();
```
yoni206, line 142: Can you assert something about `len_res_even` and `les_res_odd`? Maybe you know one if even and one is odd?

src/sorting_network.cpp
```diff
@@ -0,0 +1,199 @@
...
+      res.push_back(two_sorted[1]);
+    }
+  }
+  else if (!sorted1_is_even && !sorted2_is_even)
+  {
+    Term first_res_odd = res_odd[0];
+    Term last_res_odd = res_odd.back();
+
+    res.push_back(first_res_odd);
+    for (size_t i = 0; i < len_res_even; ++i)
```
yoni206, line 180: Can you add documentation about these cases?

tests/test-sorting-network.cpp
```diff
@@ -0,0 +1,103 @@
...
+        << "Incorrect sorting network for model " + model;
+
+    solver->pop();
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    ParameterizedSolverSortingNetworkTests,
+    SortingNetworkTests,
+    testing::ValuesIn(available_non_generic_solver_configurations()));
```
yoni206, line 101: why can't we test this with the generic solver?

tests/test-sorting-network.cpp
```diff
@@ -0,0 +1,103 @@
...
+    boolsort = solver->make_sort(BOOL);
+    for (size_t i = 0; i < NUM_VARS; ++i)
+    {
+      boolvec.push_back(solver->make_symbol("b" + std::to_string(i), boolsort));
+    }
+  }
+  SmtSolver solver;
+  Sort boolsort;
+  TermVec boolvec;
+  size_t NUM_VARS = 8;
```
yoni206, line 51: I'd suggest testing also an odd length , to test well the division by 2 of the pivot. Maybe even something like `3` -- because when u divide it by 2 u get `1` and then `0`, which seem like edge cases.

tests/test-sorting-network.cpp
```diff
@@ -0,0 +1,103 @@
...
+};
+
+TEST_P(SortingNetworkTests, TestSortingNetwork)
+{
+  SortingNetwork sn(solver);
+  TermVec sorted = sn.sorting_network(boolvec);
+
+  // Test each possible return value
+  for (size_t num_true = 0; num_true <= NUM_VARS; ++num_true)
```
yoni206, line 60: Cool test! :)

tests/test-sorting-network.cpp
```diff
@@ -0,0 +1,103 @@
...
+    {
+      val = solver->get_value(bb);
+      if (val == true_)
+      {
+        counted_true++;
+      }
+      model += "\t" + bb->to_string() + " := " + val->to_string() + "\n";
+    }
+
+    EXPECT_EQ(num_true, counted_true)
```
yoni206, line 91: Isn't this too strong? I thought that the requirement is `>=`, not `==`.

yoni206 (changes requested):

tests/test-sorting-network.cpp
makaimann: I was going to post an issue about that soon. I keep getting failures from the generic solver that seem unrelated to the test. Maybe you could look into that?

include/sorting_network.h
makaimann: Yeah that was unclear, clarified in 03a6607a7a225e4b96a42cddd25fd335aed8df3d.

include/sorting_network.h
makaimann: Haha, added with 6fee1c02d9074ff5accb4d3094d3f37eb1cba863

include/sorting_network.h
makaimann: Good point, done in 6fee1c02d9074ff5accb4d3094d3f37eb1cba863.

src/sorting_network.cpp
makaimann: Not the best documentation, but a little better done with d6bc10dfc4f2f1fba66463bb0187347524396e1f.

tests/test-sorting-network.cpp
makaimann: Good point, added two others to cover all cases: d6bc10dfc4f2f1fba66463bb0187347524396e1f

tests/test-sorting-network.cpp
makaimann: Right, except that I'm also asserting it's not greater than. So if `y` is the sorted network, then I'm asserting `y_i` and `not y_{i+1}` which means `>=i` but not `>i`, i.e. exactly `i`.

include/sorting_network.h
```diff
@@ -0,0 +1,95 @@
...
+#include "smt.h"
+
+namespace smt {
+
+/** \class SortingNetwork
+ *         Implements a symbolic sorting network for boolean-sorted variables.
+ *         All methods are const so it can be re-used for terms from the same
+ *          solver.
+ *         Returns a vector of new terms, which can be used to determine how
+ *          many variables are assigned to true symbolically.
```
yoni206, line 27:
```suggestion
 *          many terms are assigned to true symbolically.
```

include/sorting_network.h
```diff
@@ -0,0 +1,95 @@
...
+**/
+
+#pragma once
+
+#include "smt.h"
+
+namespace smt {
+
+/** \class SortingNetwork
+ *         Implements a symbolic sorting network for boolean-sorted variables.
```
yoni206, line 23:
```suggestion
 *         Implements a symbolic sorting network for boolean-sorted terms.
```

include/sorting_network.h
```diff
@@ -0,0 +1,95 @@
...
+** \verbatim
+** Top contributors (to current version):
+**   Makai Mann
+** This file is part of the smt-switch project.
+** Copyright (c) 2020 by the authors listed in the file AUTHORS
+** in the top-level source directory) and their institutional affiliations.
+** All rights reserved.  See the file LICENSE in the top-level source
+** directory for licensing information.\endverbatim
+**
+** \brief Implements a symbolic sorting network for boolean-sorted variables.
```
yoni206, line 12:
```suggestion
** \brief Implements a symbolic sorting network for boolean-sorted terms.
```

include/sorting_network.h
yoni206: or maybe at least mention it is a helper for sorting_network_rec?

include/sorting_network.h
yoni206: ping

yoni206 (approved):
Thanks for all the changes! For your consideration before merging, I've made some more minor comments inline.

makaimann:
Great, thanks for the detailed review and lots of good catches! I committed your suggestions. I'll wait for CI before merging.

PR #239, opened 2021-05-18
Don't lex sort symbols directly in parser

makaimann:


makaimann:
Here's the comparison to `cvc4` and an old version of `smt-switch-cvc4`:

comparison-sort-symbols-noninc

comparison-sort-symbols

The errors in `qf_bv_noninc` are expected due to a known bug. Otherwise the results look good to me.

src/smtlib_reader.cpp
```diff
@@ -39,12 +39,30 @@ const unordered_map<string, vector<string>> logic_map(
...
+      { "NIA", { INT } },
+      { "NIRA", { INT, REAL } },
+      { "NRA", { REAL } },
+      { "RDL", { REAL } },
+      { "UF", { FUNCTION } } });
+
 SmtLibReader::SmtLibReader(smt::SmtSolver & solver, bool strict)
     : solver_(solver),
       strict_(strict),
+      logic_("UNSET"),
```
yoni206, line 60: Maybe document above the `logic_` field that there is a special `"UNSET"` value.

tests/smt2/qf_ufbv/test-define-sort.smt2
```diff
@@ -0,0 +1,8 @@
+(set-logic QF_UFBV)
+(define-sort Int () (_ BitVec 32))
```
yoni206, line 2: I recommend:
(a) having another less confusing example (e.g., calling it `BV`).
(b) keep this example, but also include a comment that says it tests a special case that is allowed by smtlib.

tests/smt2/qf_ufbv/test-define-sort.smt2
```diff
@@ -0,0 +1,8 @@
...
+(declare-const x Int)
+(declare-const y Int)
+(assert (= ((_ extract 31 31) x) #b1))
+(assert (= ((_ extract 31 31) y) #b0))
+(assert (bvult x y))
+(check-sat)
```
yoni206, line 8: \n

yoni206 (approved):
LGTM, some recommendations inline.

src/smtlib_reader.cpp
makaimann: Good idea! Done in 03331c979f6c972fe0924e6c471856f332ce1cab.

tests/smt2/qf_ufbv/test-define-sort.smt2
makaimann: Good idea! Done in 03331c979f6c972fe0924e6c471856f332ce1cab.

tests/smt2/qf_ufbv/test-define-sort.smt2
makaimann: Thanks!

src/smtlibparser.yy
```diff
@@ -352,29 +351,46 @@ bvconst:
...
+     if (sk == smt::NUM_SORT_KINDS)
+     {
+       // got dedicated null enum
+       yy::parser::error(@2, std::string("Unrecognized sort: ") + $2);
+       YYERROR;
+     }
+     $$ = drv.solver()->make_sort(sk, std::stoi($3));
    }
-   | SYMBOL
+   | LP SYMBOL sort sort RP
```
4tXJ7f, line 383: I may be missing something but I think that the current parser does not support sorts that are parametrized by more than two sorts and parametrized uninterpreted sorts in general. For example:

```
(declare-sort U 3)                  
(declare-const x (U Bool Bool Bool))
(check-sat)                         
```

src/smtlib_reader.cpp
```diff
@@ -100,6 +119,13 @@ void SmtLibReader::set_logic(const string & logic)
 
         processed_logic =
             processed_logic.substr(len, processed_logic.length() - len);
+
+        // should also be in logic_sortkind_map
+        assert(logic_sortkind_map.find(sub) != logic_sortkind_map.end());
+        for (const SortKind sk : logic_sortkind_map.at(sub))
+        {
+          sortkinds_[smt::to_string(sk)] = sk;
```
4tXJ7f, line 127: If we use UFs, the parser should still allow the declaration/definition of sorts named `Function`.

src/sort.cpp
```diff
@@ -24,15 +24,15 @@
...
-      { UNINTERPRETED_CONS, "UNINTERPRETED_CONS" },
-      { DATATYPE, "DATATYPE" } });
+    { { ARRAY, "Array" },
+      { BOOL, "Bool" },
+      { BV, "BitVec" },
+      { INT, "Int" },
+      { REAL, "Real" },
+      { FUNCTION, "Function" },
+      { UNINTERPRETED, "Uninterpreted" },
+      { UNINTERPRETED_CONS, "UninterpretedSortConstructor" },
```
4tXJ7f, line 34: Out of curiosity: What is the purpose of those two kinds?

4tXJ7f (changes requested):
I think that the changes make sense but I have some minor comments.

src/smtlibparser.yy
makaimann: That's right. It doesn't support `declare-sort` at all yet actually. I'll add that to the list of features to add. Thanks for pointing that out!

src/sort.cpp
makaimann: Good question, it might not be the best solution. Essentially it distinguishes between an uninterpreted sort, and a parameterized uninterpreted sort before it has been instantiated. So if we had
```
(declare-sort U 3)
```
Then `U` would have `SortKind` `UNINTERPRETED_CONS`, but `(U Bool Bool Bool)` would have `SortKind` `UNINTERPRETED`.

src/smtlib_reader.cpp
makaimann: Good point! Fixed that with fa2f52bb52fcbbe1a65543ce1ec06bb9e30be37e.

PR #245, opened 2021-05-27
Add a contribution policy

makaimann:
This PR would add an external contribution policy and adds a THANKS file, both based on those used in [cvc5](https://github.com/cvc5/cvc5).


barrettcw (approved):
LGTM

makaimann:
Thanks for the review!

PR #248, opened 2021-06-09
Add basic support for attributes

makaimann:
This PR adds SMT-LIB attributes to the grammar. However, the basic `SmtLibReader` implementation does nothing with them but print a warning that they're being ignored.


include/smtlib_reader.h
```diff
@@ -150,6 +150,17 @@ class SmtLibReader
    */
   virtual void pop(uint64_t num = 1);
 
+  /** Mark a term with an attribute
+   *  This currently is unimplemented and just prints a warning.
```
yoni206, line 154: Unimplemented? Or is it implemented but the attribute is ignored?

include/smtlib_reader.h
```diff
@@ -150,6 +150,17 @@ class SmtLibReader
...
+   *  It can be overridden in a derived class to do something
+   *  useful with the attribute
+   *  @param sym the symbol
+   *  @param term the term
```
yoni206, line 158: Can you update these?

src/smtlibparser.yy
```diff
@@ -292,19 +302,29 @@ s_expr:
...
+    LP let_term_bindings RP term_s_expr RP
   {
     drv.pop_scope();
     $$ = $7;
   }
+  | LP EP term_s_expr attribute RP
+  {
+    // the default implementation does nothing
+    // but print a warning to standard error
+    // can implement the function in derived class
```
yoni206, line 314:
```suggestion
    // but print a warning to standard error.
    // it is possible to implement the function in derived class
```

tests/smt2/qf_ufbv/test-attr.smt2
```diff
@@ -0,0 +1,16 @@
...
+(define-fun ylow () (_ BitVec 4) ((_ extract 3 0) y))
+(define-fun xhigh () (_ BitVec 4) ((_ extract 7 4) x))
+(define-fun yhigh () (_ BitVec 4) ((_ extract 7 4) y))
+(declare-fun f ((_ BitVec 8)) (_ BitVec 8))
+(assert (! (distinct (f x) (f y)) :name diff))
+(assert (! (= xlow ylow) :name equal-low))
+(check-sat)
+(assert (! (= xhigh yhigh) :name equal-high))
+(check-sat)
+(exit)
```
yoni206, line 16:
```suggestion
(exit)

```

tests/smt2/qf_ufbv/test-attr.smt2
```diff
@@ -0,0 +1,16 @@
...
+(set-option :incremental true)
+(declare-const x (_ BitVec 8))
+(declare-const y (_ BitVec 8))
+(define-fun xlow () (_ BitVec 4) ((_ extract 3 0) x))
+(define-fun ylow () (_ BitVec 4) ((_ extract 3 0) y))
+(define-fun xhigh () (_ BitVec 4) ((_ extract 7 4) x))
+(define-fun yhigh () (_ BitVec 4) ((_ extract 7 4) y))
+(declare-fun f ((_ BitVec 8)) (_ BitVec 8))
+(assert (! (distinct (f x) (f y)) :name diff))
+(assert (! (= xlow ylow) :name equal-low))
```
yoni206, line 12: Is this iteration supposed to support arbitrary attributes? If so, perhaps we should add more examples to this test, like quantifier patterns, that use terms and not only strings. 
Perhaps we can just use this example from `cvc5`'s regressions as is:
https://github.com/cvc5/cvc5/blob/master/test/regress/regress0/quantifiers/double-pattern.smt2

yoni206 (approved):
LGTM modulo minor comments and suggestions.

tests/smt2/qf_ufbv/test-attr.smt2
makaimann: Ahh interesting, I was wondering what `sexprs` in attributes were for. So, this should currently parse but I kept it as a string. So in this case it would get the string literal `(P x)`, as opposed to the term `P` applied to `x`. Do you think I should interpret it as a term in the parser and have a `term_attribute` method that takes a term instead of a keyword and value pair?

include/smtlib_reader.h
makaimann: Good point, fixed in 793be8ca80bf09cf834920d55f3564cdc93cdb64.

include/smtlib_reader.h
makaimann: Oops, thanks for catching that! Fixed in 793be8ca80bf09cf834920d55f3564cdc93cdb64.

tests/smt2/qf_ufbv/test-attr.smt2
makaimann: Hey, I know this has already been approved but I wanted to revisit this discussion. Also interested in feedback from @4tXJ7f. Should s-expr in attributes be processed as terms instead of strings?

Thanks!

makaimann:
I'm going to merge this now, but we can revisit term attributes for quantifier triggers and things like that in a subsequent PR.

PR #250, opened 2021-06-11
initial_dimacs

BhaskarGupta22:
Added cnf_to_dimacs function in btor_qf_ufbv.cpp in examples directory(to convert a cnf to DIMACS format).


yoni206 (changes requested):
Alright, great, and thanks!

First request before I actually review -- the diff is very big:
image

Seems like you added many new files to the repository (maybe you used `git add -A`?)
The only new files should be the file where you implemented the new procedure and the test files. For example, build files should not be added.

Please try using something like `git rm` to remove the redundant additions.

yoni206:
@makaimann , just FYI and feel free to give feedback also -- @BhaskarGupta22 is adding initial support for SAT back ends by dumping DIMACS.

BhaskarGupta22:
Removed the build files

yoni206:
OK, that is much better :)
image


Now, let's start by moving this to a proper place and also the test to a proper place. 
I suggest adding this functionality to https://github.com/makaimann/smt-switch/blob/master/src/utils.cpp (and the corresponding declaration to the h file).

Regarding the test, I suggest removing it from the implementation and adding it here:
https://github.com/makaimann/smt-switch/blob/master/tests/unit/unit-util.cpp


Further, all changes to `CMakelists.txt` files *should* be added to the repository, as otherwise it will not get built.

BhaskarGupta22:
I'll add the functionality to utils.cpp and the declaration to utils.h

Should I assert string equivalence in the test or is there any better you to test it?

I have used build.sh in the examples directory to build and have not made any changes to CMakelists.txt by my own.

makaimann:
This sounds like a great feature, thanks for working on this! FYI, I [approved the workflow runs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) so the CI tests will run.

examples/btor_qf_ufbv.cpp
```diff
@@ -51,4 +51,4 @@ int main()
  UnorderedTermSet ua;
  s->get_unsat_assumptions(ua);
  for (Term t : ua) { cout << t << endl; }
-}
+}
```
yoni206, line 54: Github complaints if there is no newline at the end.

```suggestion
}

```

src/utils.cpp
```diff
@@ -195,6 +196,104 @@ bool is_lit(const Term & l, const Sort & boolsort)
   return false;
 }
 
+
+
+std::string cnf_to_dimacs(Term cnf){
```
yoni206, line 201: Let's have this function take a second argument, similar to this function:
https://github.com/makaimann/smt-switch/blob/ed05b58d109642f440e6311c632131cdeb20991a/include/printing_solver.h#L45

instead of returning a string.

The string can be then obtained from the stream for testing.

src/utils.cpp
```diff
@@ -195,6 +196,104 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+   TermVec vecs({cnf});
```
yoni206, line 202: Please add more comments throughout the code that explain what it is doing.

src/utils.cpp
```diff
@@ -195,6 +196,104 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+   TermVec vecs2;
```
yoni206, line 203: Please run `clang-format` on the code. It will make it look nicer, e.g., with spaces around operators etc.

tests/unit/unit-util.cpp
```diff
@@ -156,6 +156,31 @@ TEST_P(UnitUtilIntTests, Oracles)
   }
 }
 
+TEST_P(UnitUtilTests, cnf_to_dimacs)
```
yoni206, line 159: This is a great test.
Let's add some more, with some corner cases, for example:
- a formula with only one clause
- a formula with no clauses
- a clause with only one literal
- an empty clause

yoni206 (changes requested):
Great, thanks for the changes. We are definitely on the right track!
Here are some more comments about organization, formatting, and testing.

include/utils.h
```diff
@@ -98,6 +100,10 @@ void get_ops(const smt::Term & term, smt::UnorderedOpSet & out);
  */
 bool is_lit(const Term & l, const Sort & boolsort);
 
+void cnf_to_dimacs(Term cnf, std::ostringstream & y);
+
+//Returns a string in DIMACs format for a given cnf formula
```
yoni206, line 105:
```suggestion
//Returns a string in DIMACs format for a given cnf formula
void cnf_to_dimacs(Term cnf, std::ostringstream & y);
```
We usually document the function above its declaration.

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
   return false;
 }
 
+void cnf_to_dimacs(Term cnf, std::ostringstream & y)
+{
+  if (cnf->is_value() && cnf->to_string() == "true")
```
yoni206, line 204: Please also check if `cnf` is of Boolean sort. Otherwise, an exception should be thrown.

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+  {  // empty cnf formula
+    y << "p cnf 0 0\n";
+    return;
+  }
+  TermVec vecs({ cnf });
+  TermVec vecs2;
```
yoni206, line 210: Please document what is the purpose of `vecs` and `vecs2`.

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+    {
+      for (auto u : t)
+      {
+        vecs.push_back(u);
+      }
+    }
+    else
+    {
+      vecs2.push_back(t);
+    }
```
yoni206, line 231: Seems like you are assuming that the only operators are `and` and `or`.
This is good, but please assert it. You should probably also add assertions about the fact that the literals are either Boolean variables or their negations (generally speaking, literals can also be complex terms like `x=5+y`).

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+  }
+  // Storing literals from each clause, each vector in literals will contain the
+  // literals from a clause
+  std::vector<std::vector<Term>> literals;
```
yoni206, line 235: Won't `clauses` be a better name?

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+
+  for (auto u : vecs2)
+  {
+    std::vector<Term>add;
+    std::vector<Term>le({u});
```
yoni206, line 240: Please document what is the purpose of these variables. Maybe then their names will make sense ;)

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+      le.pop_back();
+      smt::Op op=t->get_op();
+      if(op.prim_op==smt::Or){
+        for(auto u:t){
+          le.push_back(u);
+        }
+      }
+      else{
+        add.push_back(t);
+      }
```
yoni206, line 254: Also around here, maybe it is a good place to add an assertion about what we assume.

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+       {  // A positive literal
+         if (ma.find(uu->to_string()) == ma.end())
+         {  // Checking if symbol is absent in the mapping done till now
+           ptr++;
+           ma[uu->to_string()] = ptr;
+         }
+       }
+       else
+       {  // A negative literal
+         Term t = (*(uu->begin()));
```
yoni206, line 280: Can you use `uu[0]`? I think this what we usually do in the code, but correct me if I am wrong.

src/utils.cpp
```diff
@@ -195,6 +199,128 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+        }
+      }
+      else{
+        add.push_back(t);
+      }
+    }
+    literals.push_back(add);
+   }
+
+   std::map<std::string, int>
```
yoni206, line 259: Why from `std::string` to `int` and not from `Term` to `int`?

tests/unit/unit-util.cpp
```diff
@@ -156,6 +156,64 @@ TEST_P(UnitUtilIntTests, Oracles)
...
+  boolsort = s->make_sort(BOOL);
+  Term a = s->make_symbol("a", boolsort);
+  Term b = s->make_symbol("b", boolsort);
+  Term c = s->make_symbol("c", boolsort);
+  Term d = s->make_symbol("d", boolsort);
+  Term clause1 = s->make_term(Or, a, s->make_term(Or, b, s->make_term(Not, c)));
+  Term clause2 = s->make_term(Or, b, s->make_term(Or, s->make_term(Not, c), d));
+  Term clause3 = s->make_term(Or, d, s->make_term(Or, s->make_term(Not, c), a));
+  Term cnf=s->make_term(And, clause1, s->make_term(And, clause2, clause3));
+//The terms in the output string is not in accordance with the order of the input because of how to function is operating on the terms
```
yoni206, line 177: Is there a way to control it? Sounds like minor changes in the function will cause this test to fail. What exactly is the problem?

tests/unit/unit-util.cpp
```diff
@@ -156,6 +156,64 @@ TEST_P(UnitUtilIntTests, Oracles)
...
+//, a dry run will show how the mapping of symbol to integer happens
+
+  // Test 1
```
yoni206, line 180: For each test, it would be useful to write in the comment which formula u r constructing.

yoni206 (changes requested):
Thanks for the changes. We are getting closer :) Some more comments below.

tests/unit/unit-util.cpp
BhaskarGupta22: (let ((.def_42 (not c))) (and (or a (or b .def_42)) (and (or b (or d .def_42)) (or d (or a .def_42)))))
Is what MSAT translated the cnf to which changed the mapping.
I tried outputting it's final string and I noticed that 'c' is being added back to the start of every clause and so it is being mapped to 1. I am thinking of only letting CVC4 test this function as it doesn't fidget with the formula and as I am using string assertion. Later when we use this formula for some other purpose we might assert something else so it won't be a problem.

tests/unit/unit-util.cpp
BhaskarGupta22: Will do that

src/utils.cpp
BhaskarGupta22: I had a feeling that mapping std::string would lead to less overhead than 'Term', though I might be wrong. Is there any disadvantage in using std::string?

src/utils.cpp
BhaskarGupta22: I tried uu[0] right now and got a compilation error, I picked this notation from is_lit() function in the same file :)

src/utils.cpp
BhaskarGupta22: I'll do that

src/utils.cpp
BhaskarGupta22: each vector will finally contain a broken-up clause, so I thought that the vector now contains literals instead of clauses.

yoni206: but this variable is the vector of these vectors, that is, the "internal" vectors could have been called "literals", but not the entire vector of vector? Or I am missing something

src/utils.cpp
yoni206: :) Well I guess we don't usually use it in the code then!

src/utils.cpp
BhaskarGupta22: Yes, this seems more appropriate.

src/utils.cpp
yoni206: Since the strings are just variable names, I don't think there will be a noticeable difference in the overhead. But even if there is, I expect the overhead of storing terms will be smaller, because probably the most naive hashing function that solvers can implement is transforming to a string and then compute the hash. But some solvers do much better than that.

The disadvantage I see in using `string` is the extra level of indirection. I think what we are really interested in this map are the terms, their string representation is somewhat irrelevant, unless I am missing its relevancy.

tests/unit/unit-util.cpp
yoni206: oh yes. So, maybe the best thing is to filter to only use "logging solvers" at the bottom of the test. This will ensure no rewriting or re-ordering is done.

src/utils.cpp
BhaskarGupta22: Ok, I'll change this

BhaskarGupta22: Btw I had first thought of just entering the terms in a vector, sorting them and them using binary_search on the terms to find their index, but it seemed like this function would be the least expensive computationally and mapping seemed the shortest to code

yoni206:
Yes I think mapping is the best option. And assuming an efficient hashing
of `Term`s it is also more efficient. But if we ever see a bottleneck here
we can revisit.

src/utils.cpp
```diff
@@ -195,6 +199,140 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+  // smt::and op is not detected that term is added back to before_and_elimination, and as no term
+  // with smt::or as the primOp is touched all clauses shall be separated and be
+  // intact
+  while (!before_and_elimination.empty())
+  {
+    Term t = before_and_elimination.back();
+    before_and_elimination.pop_back();
+    smt::Op op = t->get_op();
+    std::string ops = op.to_string();
+    assert(ops=="null" || ops=="or" || ops=="and" || ops=="not");
```
yoni206, line 224: No reason to use strings here, right? You can do it similarly to the line below.

src/utils.cpp
```diff
@@ -195,6 +199,140 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+    std::vector<Term>after_or_elimination;
+    std::vector<Term>before_or_elimination({ u });
+    while (!before_or_elimination.empty())
+    {  // This while loop functions in the same way as above and eliminates
+       // smt::or by separating the literals
+      Term t = before_or_elimination.back();
+      before_or_elimination.pop_back();
+      smt::Op op = t->get_op();
+      std::string ops = op.to_string();
+      assert(ops=="null" || ops=="or" || ops=="not");
```
yoni206, line 252: Same as above, no need for strings.

src/utils.cpp
```diff
@@ -195,6 +199,140 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+    }
+    clauses.push_back(after_or_elimination);
+   }
+
+   std::map<Term, int>
+       ma;  // This map will create a mapping from symbols to distinct
+            // contiguous integer values.
+   int ptr = 0;  // pointer to store the next integer used in mapping
+
+   for(auto u : clauses)
```
yoni206, line 275: Please document above this line what this loop is doing.

src/utils.cpp
```diff
@@ -195,6 +199,140 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+   y << "p cnf ";
+   y << ptr;  // number of distinct symbols
+   y << " ";
+
+   int sz = clauses.size();
+
+   y << sz;  // number of clauses
+   y << "\n";
+
+   for (auto u : clauses)
```
yoni206, line 311: Please document above this line what this loop is doing.

yoni206 (changes requested):
Thanks for the changes, more comments inline.
In addition:
1. Please run the output of your tests on some SAT solver (e.g. CaDiCal) to make sure it works correctly.
2. Please run `clang-format` (and not `git clang-format`) on the entire file. I still see some formatting issues.

tests/unit/unit-util.cpp
```diff
@@ -156,6 +156,64 @@ TEST_P(UnitUtilIntTests, Oracles)
   }
 }
 
+TEST_P(UnitUtilTests, cnf_to_dimacs)
+{
+  if (s->get_solver_enum() != CVC4)
+  {
+    return;
+  }
```
yoni206, line 164: I think we do want to make sure this utility can be used outside of `cvc4` and also tested.
So please try to do something similar to this: 
https://github.com/makaimann/smt-switch/blob/master/tests/test-logging-solver.cpp

They only create Logging Solvers. These are guaranteed not to change the structure of the input formulas. And they do this for every supported solver.
You might want to create a separate class in this file for the dimacs tests, and for that class, iterate (using the google tests mechanism) over only logging solvers.

BhaskarGupta22: Should I transfer my test from the earlier file to this file?

BhaskarGupta22: I don't think I exactly understood what I should do

yoni206: Good question!
Your current test file ends with these two commands:
https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR218
https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR222

The former controls the executions of the tests that are defined with `UnitUtilTests`, and the latter relates to `UnitUtilIntTests`.
In the former, the command tells google test to execute all tests related to `UnitUtilTests` on all solvers that support `TERMITER`.
In the latter, a more restricted set of solvers is invoked -- those that support both `TERMITER` and `THEORY_INT`.

Now, `UnitUtilTests` and `UnitUtilIntTests` are defined [here](https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR31) and [here](https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR50).

Then, each test is associated with one of them. For example, [this](https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR85) test is associated with the former.

What I am suggesting is to add a third class, call it `UnitUtilDimacsTests`. It might contain only one test, but that's OK. Maybe we will add more tests to it later.
So your current test should be associated with this new class. 
Then, you can add a line at the end of the file, just like [this](https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR218) line.

But, instead of using `filter_solver_configurations`, use `filter_solver_enums`. A `SolverConfiguration` is a solver enum + a flag that states whether it is a logging solver or not. For this test, we only want logging solvers. So we just range over the enums.
Then, in your new `UnitUtilDimacsTests` class, the setup line will not be like it is now:
https://github.com/makaimann/smt-switch/pull/250/files#diff-77a1d128c2d18757a0ac996385ff8f6ebc896eb2abf3c83647bdf1ee35124d3dR37

Because, now, `GetParams` will return a `SolverEnum`, while `create_solver` expects a `SolverConfiguration`.
So create a `SolverConfiguration` that takes the enum, and give it ` true` flag for the `logging` argument. This will ensure you will get a logging solver, which in turn, will ensure that the original assertions are not rewritten before you process them.

This was a long answer, I hope it is not too confusing :) Let me know if you have more questions. There might be an easier way to achieve the same goal of course, but this is how I would do it.

BhaskarGupta22: Thanks a lot! Gained a lot of clarity.

tests/unit/unit-util.cpp
```diff
@@ -47,6 +47,28 @@ class UnitUtilTests : public ::testing::Test,
...
+
+    boolsort = s->make_sort(BOOL);
+    for (size_t i = 0; i < 30; ++i)
+    {
+      symbols.push_back(s->make_symbol("x" + std::to_string(i), boolsort));
+    }
+  }
+  SmtSolver s;
+  Sort boolsort;
+  TermVec symbols;
```
yoni206, line 69: What are these for?

tests/unit/unit-util.cpp
```diff
@@ -156,6 +178,60 @@ TEST_P(UnitUtilIntTests, Oracles)
   }
 }
 
+TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
+{
+  boolsort = s->make_sort(BOOL);
```
yoni206, line 183: You already have this as a class member, no need to redefine it.

tests/unit/unit-util.cpp
yoni206: Is this still relevant?

tests/unit/unit-util.cpp
```diff
@@ -47,6 +47,28 @@ class UnitUtilTests : public ::testing::Test,
...
 
+class UnitUtilDimacsTests
+    : public ::testing::Test,
+      public ::testing::WithParamInterface<SolverConfiguration>
+{
+ protected:
+  void SetUp() override
+  {
+    SolverConfiguration se = GetParam();
+    SolverConfiguration sc(se.solver_enum, true);
```
yoni206, line 58: Here you can explain that we are only using logging solvers for this test because we don't want the original formula to change -- otherwise it might no longer be a CNF.

tests/unit/unit-util.cpp
```diff
@@ -165,4 +241,9 @@ INSTANTIATE_TEST_SUITE_P(ParameterizedUnitUtilIntTests,
                          UnitUtilIntTests,
                          testing::ValuesIn(filter_solver_configurations({ TERMITER, THEORY_INT })));
 
+INSTANTIATE_TEST_SUITE_P(
+    ParameterizedUnitUtilDimacsTests,
+    UnitUtilDimacsTests,
+    testing::ValuesIn(filter_solver_configurations({ TERMITER, THEORY_INT })));
```
yoni206, line 247: I don't see a reason u would need support for `THEORY_INT` for this test. And in fact, possibly also no need for TERMITER. So maybe just use `available_solver_enums`.

yoni206 (changes requested):

tests/unit/unit-util.cpp
BhaskarGupta22: I removed TermVec, if I don't mention the other two the compiler throws an error of "undeclared variable"

tests/unit/unit-util.cpp
BhaskarGupta22: I am unable to reply to the comment above this one, so I am replying here. My function is taking the last element of the vector during "and elimination", recursively removing and pushing it to the end of a new vector, due to which the order of clauses change, a similar thing occurs during "or elimination". The set of clauses, in the end, are the same, so I don't think it would matter much.

tests/unit/unit-util.cpp
```diff
@@ -165,4 +237,8 @@ INSTANTIATE_TEST_SUITE_P(ParameterizedUnitUtilIntTests,
                          UnitUtilIntTests,
                          testing::ValuesIn(filter_solver_configurations({ TERMITER, THEORY_INT })));
 
+INSTANTIATE_TEST_SUITE_P(ParameterizedUnitUtilDimacsTests,
+                         UnitUtilDimacsTests,
+                         testing::ValuesIn(available_solver_configurations()));
```
yoni206, line 242: better to use `available_solver_enums` -- currently u r running each test twice.

tests/unit/unit-util.cpp
yoni206: yes, because you are using them! the google tests framework makes these class fields available in the test bodies.

src/utils.cpp
```diff
@@ -195,6 +199,141 @@ bool is_lit(const Term & l, const Sort & boolsort)
...
+    return;
+  }
+  TermVec before_and_elimination({ cnf });
+  TermVec after_and_elimination;
+  // This while loop separate the clauses, after_and_elimination will contain
+  // all clauses because every smt::and op will be eliminated. This happens
+  // because until smt::and op is not detected that term is added back to
+  // before_and_elimination, and as no term with smt::or as the primOp is
+  // touched all clauses shall be separated and be intact
+  while (!before_and_elimination.empty())
```
BhaskarGupta22, line 218: The recursive step implemented iteratively

tests/unit/unit-util.cpp
BhaskarGupta22: Yes, just realized now on seeing the code that both true and false are pushed for each enum.

yoni206 (approved):
Thanks for this and for all the changes!

yoni206 (approved):

PR #251, opened 2021-06-25
Generic Datatypes Sorts

afnaanhashmi:
Encodes most functionality for the creation and usage of generic sorts that pertain to inductive datatypes. Final sticking point: keeping generic_sort.cpp light enough to work with multiple datatypes. As of right now, there is no way to distinguish generic sorts of sortkind DATATYPE. This can be remedied with an unordered_map and a new make_generic_sort() function, but I'm unsure of how bulky we want the generic sort files to be.


src/generic_solver.cpp
```diff
@@ -273,6 +279,7 @@ string GenericSolver::run_command(string cmd, bool verify_success_flag) const
   write_internal(cmd);
   // reading the result
   string result = read_internal();
+  //cout << "Result : \n" + result <<endl;
```
yoni206, line 282: please remove, as we don't want this into master.

src/generic_solver.cpp
```diff
@@ -478,31 +485,81 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
 
 Sort GenericSolver::make_sort(const DatatypeDecl & d) const
 {
-  throw NotImplementedException("Generic Solvers do not support datatypes");
+  std::string dtName = (*datatypedecl_name_map)[d];
```
yoni206, line 488: Before this line, please add an assertion that `d` exists in the map. It is very useful, because in case of a bug these assertions help understand what's going on. Also, if `d` happens to not exist in the map, this line will add it with some default value, and we don't want that.

ere is an example for such an assertion:
https://github.com/makaimann/smt-switch/pull/251/files#diff-08c7035034a6ef997827cf24e4181dd3e3de5e1ebc1ebe19c220dd92ec52624fR71

src/generic_solver.cpp
```diff
@@ -478,31 +485,81 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
...
+  if (name_sort_map->find(dtName) == name_sort_map->end()) {
+    std::string toSolver = "(" + DECLARE_DATATYPE_STR + " ((";
```
yoni206, line 490: Let's keep variable names with lower case and underscores. Same for other variable names.

tests/test-dt.cpp
```diff
@@ -46,18 +46,25 @@ TEST_P(DTTests, DatatypeDecl)
...
     DatatypeConstructorDecl consdecl = s->make_datatype_constructor_decl("cons");
     s->add_selector(consdecl, "head", s->make_sort(INT));
-    s->add_selector_self(consdecl, "tail");
-    s->add_constructor(consListSpec, nildecl);
     s->add_constructor(consListSpec, consdecl);
-    Sort listsort = s->make_sort(consListSpec);
+    //cout << (*datatypedecl_name_map)[consListSpec] << endl;
+        Sort listsort = s->make_sort(consListSpec);
+	cout << "created the sort" << endl;
+	/*
```
yoni206, line 67: Instead of commenting out, you can for now wrap this block with an `if` statement that makes sure that this is not using a generic solver. Then our gradual goal will be to remove the `if` statement, and bring the block back.

yoni206 (changes requested):
Thanks for this PR! It is a great start.
I have some initial comments inline the code.
In addition, please:
1. change the style of the code using `git clang-format`.
2. I think this is a good time to move some of the state of datatypes into a new class in a new file -- `generic_datatype.h/cpp` that would inherit from the class on `datatype.h`.

afnaanhashmi:
Thank you for the feedback! Additionally, would you recommend that I use generic_datatype.h to hold the memory needed to have multiple, uniquely named sorts of sortkind DATATYPE?

yoni206:
Let's leave it to what is more convenient to you, generally speaking it would be good to try and be consistent with `GenericSort` and `GenericTerm`.

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
 
-#pragma once
+//#pragma once
```
yoni206, line 2: why?

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
...
 
-#include <functional>
+//#include <functional>
```
yoni206, line 4: If you don't need this, it is better to remove than to comment out.

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
...
+#include "generic_sort.h"
 #include "smt_defs.h"
-#include "sort.h"
 
 // using namespace smt;
 namespace smt {
 
+struct selectorComponents
+{
+  std::string name;
```
yoni206, line 19: I would expect us to need to also store the corresponding constructor (every selector is associated with a constructor).

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
...
+  Sort sort;
+};
+
+class DatatypeMemoryManager
```
yoni206, line 23: I think we should try to be similar to `generic_sort.h` and `generic_term.h`. The main class of this file should be `GenericDatatype` and should inherit from `AbsDatatype`. Such a class represents a single datatype, no a global store of datatypes. More comments about this below.

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
...
+
+ private:
+  std::unique_ptr<std::unordered_map<std::string, DatatypeDecl>>
+      name_datatypedecl_map;
+  std::unique_ptr<std::unordered_map<DatatypeDecl, std::string>>
+      datatypedecl_name_map;
+  std::unique_ptr<std::unordered_map<std::string, DatatypeConstructorDecl>>
+      name_datatypeconsdecl_map;
+  std::unique_ptr<std::unordered_map<DatatypeConstructorDecl, std::string>>
+      datatypeconsdecl_name_map;
```
yoni206, line 62: These should belong to the `generic_solver.h`. Sorry if I am contradicting my past self. But the names are a part of the state of the generic solver. The generic datatype class should store the datatype details. I may have functions to compute strings, or even store its own string representation. But the class should represent a single datatype, nothing more.

include/generic_datatype.h
```diff
@@ -1,16 +1,77 @@
...
+  std::unique_ptr<
+      std::unordered_map<DatatypeDecl, std::vector<DatatypeConstructorDecl>>>
+      dtdecl_dtconsdecl_map;
+
+  std::unique_ptr<std::unordered_map<DatatypeConstructorDecl,
+                                     std::vector<selectorComponents>>>
+      dtconsdecl_selector_map;
```
yoni206, line 69: I would rather the generic datatype class having its constructors as fields, not in external maps. Similarly, a constructor class can store selectors as fields.

yoni206 (changes requested):
Thanks for this change! It is great to move some of the state out of the generic solver, but I think you moved to much outside :) Sorry that my original review was not clear enough about this. Also, to be consistent with GenericSort and GenericTerm, there should be a `GenericDatatype` class that represents a single datatypes. This is preferable than storing this information in external maps. More details in the review itself, please let me know if you have questions!

src/generic_datatype.cpp
```diff
@@ -0,0 +1,88 @@
...
+  return num_selectors;
+}
+
+void GenericDatatype::add_selector(
+    const GenericDatatypeConstructorDecl & dt_cons_decl,
+    const std::shared_ptr<selectorComponents> & newSelector)
+{
+  for (unsigned int i = 0; i < cons_decl_vector.size(); ++i)
+  {
+    if (cons_decl_vector[i].get_name() == dt_cons_decl.get_name())
```
yoni206, line 72: The fact that we chose to compare datatypes etc. by their names is an implementation detail, that will be hard to remember every time we want to compare them. So I suggest following the `AbsSort` and `GenericSort` pattern. `sort.h` overrides the `==` operator. `sort.h` implements this by calling the `compare` function, that can be implemented by various classes that inherit from `AbsSort`. In particular, `GenericDatatype` implements such a `compare` function. And when you call `==` between two generic datatypes, since `GenericDatatype` does not override the operator `==`, the code of `AbsSort` is called, and hence `compare` is executed. 

Note that this will require adding a `dataype.cpp` file.

Let me know if anything of this isn't clear!

src/generic_solver.cpp
```diff
@@ -77,7 +77,17 @@ GenericSolver::GenericSolver(string path,
...
+      datatypedecl_datatype_map(
+          new unordered_map<DatatypeDecl, std::shared_ptr<GenericDatatype>>()),
+      name_datatypedecl_map(new unordered_map<string, DatatypeDecl>()),
+      datatypedecl_name_map(new unordered_map<DatatypeDecl, string>())
+// name_datatypeconsdecl_map(new
+// unordered_map<string,DatatypeConstructorDecl>()),
+// datatypeconsdecl_name_map(new unordered_map<DatatypeConstructorDecl,
+// string>()) dtdecl_dtconsdecl_map(new unordered_map<DatatypeDecl,
+// vector<DatatypeConstructorDecl>>()), dtconsdecl_selector_map(new
+// unordered_map<DatatypeConstructorDecl, vector<selectorComponents>>)
```
yoni206, line 90: remove

src/generic_solver.cpp
```diff
@@ -106,7 +116,8 @@ GenericSolver::GenericSolver(string path,
   for (int i=0; i < read_buf_size; i++) {
     read_buf[i]=0;
   }
-
+  // solver_dt_memory = unique_ptr<DatatypeMemoryManager>(new
+  // DatatypeMemoryManager);
   // start the process with the solver binary
```
yoni206, line 121: remove

src/generic_solver.cpp
```diff
@@ -148,6 +159,7 @@ void GenericSolver::start_solver() {
     {
       argv[i] = cmd_line_args[i - 1].c_str();
     }
+    // solver_dt_memory->initialize_datatype_memory();
```
yoni206, line 162: Well, just remove all commented out lines :)

yoni206 (changes requested):
Looks much better! I think we are in the right direction now!
Some initial new comments, will be easier for me to review other parts once these are fied.
Thanks!

src/generic_datatype.cpp
afnaanhashmi: So basically, follow the model set by sort and generic sort which entails reconfiguring datatype.h and a new datatype.cpp to override the '==' function in the same way as sort.cpp, namely calling compare on the objects. And then implementing a compare function in genericdatatype (which could also just compare the names)?

include/generic_solver.h
```diff
@@ -251,6 +253,44 @@ class GenericSolver : public AbsSmtSolver
...
+    Sort sort;
+  };
+  */
+
+  // a map linking each datatype constructor declaration with a vector
+  // holding its associated selectors, in the form of the previously
+  // defined struct.
+  //   std::unique_ptr<std::unordered_map<DatatypeConstructorDecl,std::vector<selectorComponents>>>dtconsdecl_selector_map;
   // used to hash terms via their internal string representation
   std::hash<std::string> str_hash;
```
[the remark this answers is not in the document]
afnaanhashmi: I'm sorry, but move what exactly? The hash?

include/generic_solver.h
```diff
@@ -251,6 +253,44 @@ class GenericSolver : public AbsSmtSolver
   // maps between sort name and actual sort and vice verse
   std::unique_ptr<std::unordered_map<std::string, Term>> name_term_map;
   std::unique_ptr<std::unordered_map<Term, std::string>> term_name_map;
+
+  // maps between datatype declaration and name and vice versa
+  std::unique_ptr<std::unordered_map<std::string, DatatypeDecl>>
+      name_datatypedecl_map;
+  std::unique_ptr<std::unordered_map<DatatypeDecl, std::string>>
+      datatypedecl_name_map;
+  std::unique_ptr<
```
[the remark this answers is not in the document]
afnaanhashmi: You mean 'name_datatypedecl_map' and 'datatypedecl_name_map' are the redundant ones right? If so, would the additional name-datatype map even be needed considering how we could use the other map to get the datatype associated with the datatypedecl adn then from there we would have all the information we could possibly need?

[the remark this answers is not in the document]
yoni206: no, only `datatypedecl_datatype_map` seems redundant. The name maps make sense here, similar to sorts and terms.

include/generic_solver.h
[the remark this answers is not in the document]
yoni206: I think that `generic_solver.h` doesn't need `str_hash` but `generic_datatype.h` does. no?

src/generic_datatype.cpp
yoni206: yes!

include/generic_datatype.h
```diff
@@ -0,0 +1,69 @@
...
+                    const std::shared_ptr<selectorComponents> & newSelector);
+  std::vector<GenericDatatypeConstructorDecl> get_cons_vector();
+  std::string get_name() const override;
+  int get_num_constructors() const override;
+  int get_num_selectors(std::string cons) const override;
+
+ protected:
+  DatatypeDecl dt_decl;
+  std::string name;
+  std::vector<GenericDatatypeConstructorDecl> cons_decl_vector;
```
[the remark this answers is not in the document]
afnaanhashmi: Wouldn't it be simpler to simply remove the vector that would be held by the generic datatype declaration? I believe the declarations I create are basic DatatypeDecl's not GenericDatatypeDecls so it might just be more streamlined to have only one vector in the main GenericDatatypeObject.

src/generic_datatype.cpp
afnaanhashmi: As a bit of an update, changing datatype.h and cpp might affect the generic cvc4 implementation. I will know more concretely after a bit as some other lurking problemsthat I hope yo have addressed soon in my implementation might be causing the issues.

include/generic_datatype.h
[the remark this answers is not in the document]
yoni206: Sure that sounds fine.

src/generic_datatype.cpp
yoni206: I suggest that for now, just configure smt-switch without cvc4. Maybe `./configure.sh --debug --z3`. This should still test the generic solver's datatype implementation (using `z3 as a back-end instead of `cvc4`), and that way errors regarding cvc4 will disappear (for now).

If you think you can resolve this, go for it, but I wouldn't waste much time on it -- we can for sure fix that later after the generic solver works with creating datatypes sorts.

include/datatype.h
```diff
@@ -35,7 +35,8 @@ class AbsDatatypeConstructorDecl {
  public:
   AbsDatatypeConstructorDecl(){};
   virtual ~AbsDatatypeConstructorDecl(){};
-
+  // virtual std::string get_name() const = 0;
```
yoni206, line 38: plz remove all commented out code.

include/generic_datatype.h~
```diff
@@ -0,0 +1,27 @@
+
+namespace smt {
```
yoni206, line 2: plz remove emacs tmp files.
`git rm include/generic_datatype.h~`

The way to not add them is to use `git add -u` instead of `git add -A`. This will only `add` existing files to the commit, and will ignore new ones. Then we will have another problem -- you will forget to add new files. But...

include/generic_solver.h
```diff
@@ -251,8 +252,16 @@ class GenericSolver : public AbsSmtSolver
...
+
+  // maps between datatype declaration and name and vice versa
+  std::unique_ptr<std::unordered_map<std::string, DatatypeDecl>>
+      name_datatypedecl_map;
+  std::unique_ptr<std::unordered_map<DatatypeDecl, std::string>>
+      datatypedecl_name_map;
+  // Map between names and Generic datatypes
+  std::unique_ptr<
+      std::unordered_map<std::string, std::shared_ptr<GenericDatatype>>>
+      name_datatype_map;
```
yoni206, line 264: plz either have the first 2, or the third + it's converse map.

include/generic_sort.h
```diff
@@ -31,7 +31,7 @@ Sort make_generic_sort(SortKind sk, Sort sort1);
 Sort make_generic_sort(SortKind sk, Sort sort1, Sort sort2);
 Sort make_generic_sort(SortKind sk, Sort sort1, Sort sort2, Sort sort3);
 Sort make_generic_sort(SortKind sk, SortVec sorts);
-
+Sort make_generic_sort(Datatype dt, std::string & name);
```
yoni206, line 34:
```suggestion
Sort make_generic_sort(Datatype dt);
```
(and propagate to other places where appropriate)

include/generic_sort.h
```diff
@@ -41,7 +41,8 @@ std::string to_smtlib(SortKind);
 class GenericSort : public AbsSort
 {
  public:
-  GenericSort(SortKind sk); 
+  GenericSort(SortKind sk);
+  GenericSort(SortKind sk, std::string name_var);
```
yoni206, line 45: Please add a comment above this constructor, saying that it is intended only for datatype sorts.

src/generic_sort.cpp
```diff
@@ -118,10 +119,17 @@ Sort make_generic_sort(SortKind sk, SortVec sorts)
...
+Sort make_generic_sort(Datatype & dt, std::string & name)
+{
+  return make_shared<GenericDatatypeSort>(dt, name);
+}
+
 // implementations
 
 GenericSort::GenericSort(SortKind sk) : sk(sk) {}
 
+  GenericSort::GenericSort(SortKind sk, std::string name_var) : sk(sk) {}
```
yoni206, line 131:
```suggestion
  GenericSort::GenericSort(SortKind sk, std::string name_var) : sk(sk) {assert(sk == DATATYPE);}
```

src/generic_sort.cpp
```diff
@@ -170,14 +178,27 @@ string GenericSort::compute_string() const {
       }
     } else if (get_sort_kind() == SortKind::UNINTERPRETED_CONS) {
       return get_uninterpreted_name();
-    } else {
+    }
+    else if (get_sort_kind() == SortKind::DATATYPE)
+    {
+      // Placeholder response to avoid segfaulting
+      return smt::to_smtlib(SortKind::BOOL);
```
yoni206, line 185:
```suggestion
      assert(base_name != "");
      return base_name;
```

src/generic_sort.cpp
```diff
@@ -170,14 +178,27 @@ string GenericSort::compute_string() const {
...
+    }
+    else
+    {
+      cout << "right before the assert false" << endl;
```
yoni206, line 189: please remove.

src/generic_sort.cpp
```diff
@@ -228,7 +249,13 @@ bool GenericSort::compare(const Sort & s) const
...
     case DATATYPE:
     {
-      throw NotImplementedException("GenericSort::compare");
+      //throw NotImplementedException("GenericSort::compare");
+      assert(sk == DATATYPE);
+      //shared_ptr<GenericDatatypeSort> this_type_cast =
+      //reinterpret_cast<GenericDatatypeSort>(*this);
+      //shared_ptr<GenericDatatypeSort> this_type_cast = static_pointer_cast<GenericDatatypeSort>(*this);
+      shared_ptr<GenericDatatypeSort> other_type_cast = static_pointer_cast<GenericDatatypeSort>(s);
+      return base_name == other_type_cast->compute_string();
```
yoni206, line 258:
```suggestion
    // there cannot be two different datatypes with the same name.
    return to_string() == s->to_string();
```

include/generic_sort.h
```diff
@@ -187,4 +190,19 @@ class UninterpretedGenericSort : public GenericSort
...
+  GenericDatatypeSort(const Datatype & dt, std::string & name);
+  ~GenericDatatypeSort();
+  std::string get_sort_name();
+  Datatype get_datatype() const override;
+  bool compare(const Sort & s) const override;
+  std::string compute_string() const;
+
+ protected:
+  std::string sort_name;
+  Datatype  gdt;
```
yoni206, line 205:
```suggestion
  Datatype  dt;
```

include/generic_sort.h
```diff
@@ -187,4 +190,19 @@ class UninterpretedGenericSort : public GenericSort
...
+ public:
+  GenericDatatypeSort(const Datatype & dt, std::string & name);
+  ~GenericDatatypeSort();
+  std::string get_sort_name();
+  Datatype get_datatype() const override;
+  bool compare(const Sort & s) const override;
+  std::string compute_string() const;
+
+ protected:
+  std::string sort_name;
```
yoni206, line 204: plz try to eliminate this field, just use the name in the other field.

src/generic_datatype.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace smt {
+GenericDatatypeConstructorDecl::GenericDatatypeConstructorDecl(
+    const std::string & name)
+    : cons_name(name)
+{
+}
+
+void GenericDatatypeConstructorDecl::add_new_selector(
+    const shared_ptr<selectorComponents> & newSelector)
```
yoni206, line 18:
```suggestion
    const selectorComponents & newSelector)
```
(and required changes).

include/generic_datatype.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "datatype.h"
+#include "exceptions.h"
+#include "generic_sort.h"
+#include "smt_defs.h"
+
+// using namespace smt;
+namespace smt {
+
+struct selectorComponents
```
yoni206, line 12:
```suggestion
struct SelectorComponents
```

src/generic_datatype.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+int GenericDatatypeConstructorDecl::get_selector_count() const
+{
+  return selector_vector.size();
+}
+bool GenericDatatypeConstructorDecl::compare(
+    const DatatypeConstructorDecl & d) const
+{
+  // Why won't type casting like this work?
+  // GenericDatatypeConstructorDecl gdtc = (GenericDatatypeConstructorDecl) d;
```
yoni206, line 42: another reminder to remove stuff like that

src/generic_datatype.cpp
```diff
@@ -0,0 +1,101 @@
...
+GenericDatatype::GenericDatatype(const DatatypeDecl & dt_declaration,
+                                 const std::string & s)
+    : dt_decl(dt_declaration), name(s)
+{
+}
+
+void GenericDatatype::add_constructor(
+    const DatatypeConstructorDecl & dt_cons_decl)
+{
+  cons_decl_vector.push_back(dt_cons_decl);
```
yoni206, line 55: throw an exception if the constructor already exists inside the vector.

src/generic_datatype.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+void GenericDatatype::add_selector(
+    const DatatypeConstructorDecl & dt_cons_decl,
+    const std::shared_ptr<selectorComponents> & newSelector)
+{
+  for (unsigned int i = 0; i < cons_decl_vector.size(); ++i)
+  {
+    if (cons_decl_vector[i] == dt_cons_decl)
+    {
+      static_pointer_cast<GenericDatatypeConstructorDecl>(cons_decl_vector[i])->get_selector_vector().push_back(*newSelector);
```
yoni206, line 66: just call this function:
https://github.com/afnaanhashmi/smt-switch/blob/524b305eedff7cd3d3ed08c7542a9c720404deeb/include/generic_datatype.h#L35

And in the implementation of that function, throw an exception if the selector already exists.

src/generic_sort.cpp
```diff
@@ -35,7 +35,8 @@ Sort make_uninterpreted_generic_sort(Sort sort_cons,
 
 Sort make_generic_sort(SortKind sk)
 {
-  if (sk != BOOL && sk != INT && sk != REAL)
+  // Tweaked to accept datatypes temporarily
+  if (sk != BOOL && sk != INT && sk != REAL && sk != DATATYPE)
```
yoni206, line 39:
```suggestion
  if (sk != BOOL && sk != INT && sk != REAL)
```

src/solver_enums.cpp
```diff
@@ -41,7 +41,8 @@ const unordered_map<SolverEnum, unordered_set<SolverAttribute>>
             UNSAT_CORE,
             // TEMP only temporarily disabled until bitwuzla
             //      quantifier refactoring is done
-            //      see https://github.com/bitwuzla/bitwuzla/commit/605f31557ec6c635e3c617d2b0ab257309e994c4
+            //      see
+            //      https://github.com/bitwuzla/bitwuzla/commit/605f31557ec6c635e3c617d2b0ab257309e994c4
```
yoni206, line 45: plz revert this change.

tests/test-dt.cpp
```diff
@@ -46,51 +46,65 @@ TEST_P(DTTests, DatatypeDecl)
...
+      Term five = s->make_term(5, intsort);
+    }
+    */
 
     // Make datatype sort
     DatatypeDecl consListSpec = s->make_datatype_decl("list");
 
     DatatypeConstructorDecl nildecl = s->make_datatype_constructor_decl("nil");
     DatatypeConstructorDecl consdecl = s->make_datatype_constructor_decl("cons");
     s->add_selector(consdecl, "head", s->make_sort(INT));
```
yoni206, line 60: plz add somewhere in this test a comparison between datatype constructors -- 
1. two that are the same (and assert that they are the same) 
2. two that are different (and assert that they are different)

for assertions use something like `ASSERT_TRUE` or `ASSERT_FALSE` or `ASSERT_EQ` or `ASSERT_EQUAL`.

yoni206 (changes requested):

include/generic_sort.h
yoni206: Was that possible?

src/generic_datatype.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+}
+
+std::string GenericDatatypeDecl::get_name() { return dt_name; }
+
+GenericDatatypeConstructorDecl::GenericDatatypeConstructorDecl(
+    const std::string & name)
+  : cons_name(name),
+    // Should I turn this into a constant?
+    dt_name("NOT IMPLEMENTED YET")
```
yoni206, line 24: What's going on here? :)

src/generic_datatype.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+}
+
+void GenericDatatypeConstructorDecl::add_new_selector(
+    const SelectorComponents & newSelector)
+{
+  // This seems like an opportunity for a new comparison function for
+  // the struct.
```
yoni206, line 32: Comments that go into master should only clarify the code, explain variables, etc. TODO comments like that are better outside, except for special cases.

src/generic_datatype.cpp
```diff
@@ -0,0 +1,146 @@
...
+  for (unsigned int i = 0; i < selector_vector.size(); ++i)
+  {
+    if (selector_vector[i].name == (newSelector).name)
+    {
+      throw "Can't add selector. It already exists in vector!";
```
yoni206, line 37:
```suggestion
      throw "Can't add selector. It already exists in this datatype!";
```

src/generic_datatype.cpp
yoni206: And another one ;)

src/generic_datatype.cpp
```diff
@@ -0,0 +1,146 @@
...
+  return cons_name
+         == static_pointer_cast<GenericDatatypeConstructorDecl>(d)->get_name();
+}
+
+  std::string GenericDatatypeConstructorDecl::get_dt_name() const
+  {
+    return dt_name;
+  }
+
+  void GenericDatatypeConstructorDecl::change_dt_name(const std::string new_name)
```
yoni206, line 72: This is a bit scary -- why would we need that?

src/generic_sort.cpp
yoni206: +1

yoni206 (changes requested):
Thanks for the changes! Some more intermediate comments, will continue review after these.

src/generic_datatype.cpp
afnaanhashmi: This was the first approach I thought of to give a value to a protected member after initialization. Because we only know the dt_name/datatype when we are actually adding a constructor to a datatype, I needed some way of updating the member. Would you recommend a particular approach to this task because this problem persists after transitioning over to using a datatype member instead.

yoni206: There is the name of the constructor, and the name of the datatype -- these are different -- don't you know the name of the datatype already when you create a `DatatypeDecl`?

afnaanhashmi: Yes we have access to the datatype name through each datatypedecl. My problem is to have a member variable in the constructor class that keeps track of the datatype/datatypedecl itself. As far as I can tell, we need a way to change the value of that member variable after initializaiton.

yoni206: But why would the datatype/datatypedecl's name will change after you add a constructor to it?

afnaanhashmi: I see now that the function name is misleading. I never intended to change the datatype's name with the function, merely the member called "dt_name" that the constructor class holds. This function was only intended to be called once when a constructor is added to a datatype so that the constructor has a record of the name of the datatype its associated with.

yoni206: Let's take this particular issue offline tomorrow :)

afnaanhashmi: I see now that the function name is misleading. I never intended to change the datatype's name with the function, merely the member called "dt_name" that the constructor class holds. This function was only intended to be called once when a constructor is added to a datatype so that the constructor has a record of the name of the datatype its associated with.

src/generic_solver.cpp
```diff
@@ -478,31 +481,121 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
...
+      to_solver += ")";
+    }
+    to_solver += ")\n))";
+    // Exact functionality of make_genericsort, without the linking
+    // errors (undefined reference when I call the new
+    // make_generic_sort).
+    assert(name_sort_map->find(dt_decl_name) == name_sort_map->end());
+    (*name_sort_map)[dt_decl_name] = dt_sort;
+    (*sort_name_map)[dt_sort] = dt_decl_name;
+    cout << to_solver << endl;
```
yoni206, line 537: please remove.

src/generic_solver.cpp
```diff
@@ -478,31 +481,121 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
...
   {
-    throw NotImplementedException("Generic Solvers do not support datatypes");
+    shared_ptr<SelectorComponents> newSelector =
+        make_shared<SelectorComponents>();
+    shared_ptr<GenericDatatypeConstructorDecl> gdt_cons = static_pointer_cast<GenericDatatypeConstructorDecl>(dt);
+    string dt_decl_name = gdt_cons->get_dt_name();
+    
+    newSelector->name = name;
+    newSelector->sort = make_shared<GenericSort>(name);
+    //assert(name_datatype_map->find(dt_decl_name) != name_datatype_map->end());
```
yoni206, line 596: please remove, or, if possible, uncomment.

include/generic_datatype.h
```diff
@@ -0,0 +1,75 @@
...
+#include "generic_sort.h"
+
+// using namespace smt;
+namespace smt {
+
+struct SelectorComponents
+{
+  std::string name;
+  Sort sort;
+  bool finalized;
```
yoni206, line 17: plz add documentation to this component.

include/generic_sort.h
```diff
@@ -41,7 +45,11 @@ std::string to_smtlib(SortKind);
 class GenericSort : public AbsSort
 {
  public:
-  GenericSort(SortKind sk); 
+  GenericSort(SortKind sk);
+  // This second constructor is only for datatype sorts
+  GenericSort(SortKind sk, std::string sort_name);
```
yoni206, line 50: In the line above, please remove the word "second" -- I am not sure if it will always stay the second constructor, but I am sure that if it doesn't we will forget to update the comment :)

include/generic_sort.h
```diff
@@ -41,7 +45,11 @@ std::string to_smtlib(SortKind);
...
+  // ONly for placeholder sorts
```
yoni206, line 51: Please mention that these are used for datatypes.

include/generic_datatype.h
```diff
@@ -0,0 +1,75 @@
...
+  GenericDatatype(const DatatypeDecl & dt_declaration);
+  virtual ~GenericDatatype(){};
+  void add_constructor(const DatatypeConstructorDecl & dt_cons_decl);
+  void add_selector(const DatatypeConstructorDecl & dt_cons_decl,
+                    const SelectorComponents & newSelector);
+  std::vector<DatatypeConstructorDecl> get_cons_vector();
+  std::string get_name() const override;
+  int get_num_constructors() const override;
+  int get_num_selectors(std::string cons) const override;
+  void change_sort_of_selector(const Sort new_sort);
```
yoni206, line 65: Please document every non-override function, especially this one.

src/generic_datatype.cpp
```diff
@@ -0,0 +1,149 @@
...
+  {
+    for (unsigned int i = 0; i < cons_decl_vector.size(); ++i) {
+      std::shared_ptr<GenericDatatypeConstructorDecl> cons_cast = static_pointer_cast<GenericDatatypeConstructorDecl>(cons_decl_vector[i]);
+      for (unsigned int f = 0; f < get_num_selectors(cons_cast->get_name()); ++f) {
+	
+	if (cons_cast->selector_vector[f].finalized == false) {
+	  cons_cast->selector_vector[f].sort = new_sort;
+	  cons_cast->selector_vector[f].finalized = true;
+	}
+      }
```
yoni206, line 144: please clang-format everything. There are several places that are not formatted, so maybe do an overall format. Let me know if you need help with that.

src/generic_solver.cpp
```diff
@@ -478,31 +481,121 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
...
+        make_shared<SelectorComponents>();
+    shared_ptr<GenericDatatypeConstructorDecl> gdt_cons = static_pointer_cast<GenericDatatypeConstructorDecl>(dt);
+    string dt_decl_name = gdt_cons->get_dt_name();
+    
+    newSelector->name = name;
+    newSelector->sort = make_shared<GenericSort>(name);
+    newSelector->finalized = false;
+    assert(name_datatype_map->find(dt_decl_name) != name_datatype_map->end());
+    shared_ptr<GenericDatatype> curr_dt = (*name_datatype_map)[dt_decl_name];
+    gdt_cons->add_new_selector(*newSelector);
```
yoni206, line 598: Please add documentation explaining your code, generally, but here in particular because we are doing something special here with the placeholder sort.

src/generic_sort.cpp
```diff
@@ -118,9 +121,19 @@ Sort make_generic_sort(SortKind sk, SortVec sorts)
...
+}
+
 // implementations
 
-GenericSort::GenericSort(SortKind sk) : sk(sk) {}
+  GenericSort::GenericSort(SortKind sk) : sk(sk) {}
+
+  GenericSort::GenericSort(SortKind sk, std::string sort_name) : sk(sk) {}
+
+  // Only used to make placeholders
```
yoni206, line 135: Please elaborate here, to make this comment clearer.

src/generic_sort.cpp
```diff
@@ -323,5 +347,32 @@ SortVec UninterpretedGenericSort::get_uninterpreted_param_sorts() const
   return param_sorts;
 }
 
+GenericDatatypeSort::GenericDatatypeSort(const Datatype & dt)
+    : GenericSort(DATATYPE),
+      gdt(dt)
+      //GenericSort::base_name((static_pointer_cast<GenericDatatype>(dt))->get_name())
```
yoni206, line 353: remove commented out lines :) not the only place currently

yoni206 (commented):
Thanks for the changes! Getting closer :)
General comments:
1. Please make sure that all the code is formatted using clang-format. Let me know if you need help with that.
2. Please add documentation to the code. Every function declaration in an h file that does not override an existing function should be documented (I guess getters and setters don't need to be). And in cpp files, try to add documentation on what the code does, espcially regarding special decisions that we have made (like placeholder sorts) but not only.
3. Please remove all commented out lines.

Thanks!

CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+
```
yoni206, line 9:
```suggestion
```

include/generic_datatype.h
```diff
@@ -0,0 +1,88 @@
...
+  std::string dt_name;
+};
+
+class GenericDatatypeConstructorDecl : public AbsDatatypeConstructorDecl
+{
+ public:
+  GenericDatatypeConstructorDecl(const std::string & name);
+  virtual ~GenericDatatypeConstructorDecl(){};
+  // Stores a new selector in the constructor object. newSelector: the
+  // selectorComponents to be added.
```
yoni206, line 41:
```suggestion
  // SelectorComponents to be added.
```

include/generic_datatype.h
```diff
@@ -0,0 +1,88 @@
...
+  void add_new_selector(const SelectorComponents & newSelector);
+  std::vector<SelectorComponents> get_selector_vector();
+  std::string get_name() const;
+  // Returns the size of selector_vector
+  int get_selector_count() const;
+  bool compare(const DatatypeConstructorDecl & d) const override;
+  std::string get_dt_name() const;
+  // Setter for the dt_decl member. Only to be used when a constructor
+  // is added to a datatype. datatype_decl: The DatatypeDecl of the datatype.
+  void update_stored_dt(const DatatypeDecl & datatype_decl);
```
yoni206, line 51: This can be `protected`, right?

include/generic_datatype.h
```diff
@@ -0,0 +1,88 @@
...
+  // Stores a new selector (newSelector) in the constructor object
+  // (dt_cons_decl) if the constructor is associated with the datatype
+  void add_selector(const DatatypeConstructorDecl & dt_cons_decl,
+                    const SelectorComponents & newSelector);
+  std::vector<DatatypeConstructorDecl> get_cons_vector();
+  std::string get_name() const override;
+  int get_num_constructors() const override;
+  int get_num_selectors(std::string cons) const override;
+  // Updates the sort of any selector whose finalized field is
+  // false. The not-finalized selectors have their sorts set to new_sort.
```
yoni206, line 77: Great, please explain in more detail here why do we do this, even put an example.

include/generic_sort.h
```diff
@@ -41,7 +45,12 @@ std::string to_smtlib(SortKind);
 class GenericSort : public AbsSort
 {
  public:
-  GenericSort(SortKind sk); 
+  GenericSort(SortKind sk);
+  // This constructor is only for datatype sorts.
+  GenericSort(SortKind sk, std::string sort_name);
+  // Only for placeholder sorts used for inductive datatypes. name
```
yoni206, line 51:
```suggestion
  // Only for placeholder sorts used for datatypes. name
```
`Inductive`is sometimes used to refer to a subset of datatypes, but here we cover also non-inductive datatypes (like records).

include/generic_sort.h
```diff
@@ -41,7 +45,12 @@ std::string to_smtlib(SortKind);
 class GenericSort : public AbsSort
 {
  public:
-  GenericSort(SortKind sk); 
+  GenericSort(SortKind sk);
+  // This constructor is only for datatype sorts.
```
yoni206, line 49: Is this constructor ever used? If not, please delete.

include/generic_sort.h
```diff
@@ -110,11 +119,13 @@ class GenericSort : public AbsSort
...
   std::string to_string() const override;
 
+  std::string get_base_name() const;
+
  protected:
   // internal function to compute
   // the string representation of a sort
   std::string compute_string() const;
-
+  std::string base_name;
```
yoni206, line 128: OK maybe now that the PR is very mature, it is time to get rid of this field :) Basically, you should only need it in `GenericDatatypeSort`, and there you have a `Datatype` that you can down-cast to `GenericDatatype`, which has a `GenericDatatypeDecl`, which has a name.

Let me know where it goes wrong.

include/generic_sort.h
```diff
@@ -187,4 +198,20 @@ class UninterpretedGenericSort : public GenericSort
...
+class GenericDatatypeSort : public GenericSort
+{
+ public:
+  GenericDatatypeSort(const Datatype & dt);
+  ~GenericDatatypeSort();
+  // Returns the name of the datatype
+  std::string get_sort_name();
+  Datatype get_datatype() const override;
+  bool compare(const Sort & s) const override;
+  std::string compute_string() const;
```
yoni206, line 210: How come thid doesn't `override`?

include/generic_sort.h
```diff
@@ -187,4 +198,20 @@ class UninterpretedGenericSort : public GenericSort
   SortVec param_sorts;  
 };
 
+class GenericDatatypeSort : public GenericSort
+{
+ public:
+  GenericDatatypeSort(const Datatype & dt);
+  ~GenericDatatypeSort();
+  // Returns the name of the datatype
+  std::string get_sort_name();
```
yoni206, line 207: Hopefully, this won't be needed either, because you could use `to_string`?

src/generic_datatype.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // If the constructor is associated with the datatype
+    if (cons_decl_vector[i] == dt_cons_decl)
+    {
+      // Adds the selector to the correct constructor
+      static_pointer_cast<GenericDatatypeConstructorDecl>(cons_decl_vector[i])
+          ->add_new_selector(newSelector);
+    }
+    else
+    {
+      throw "Can't add selector. The constructor is not a member of the datatype!";
```
yoni206, line 110: Great! Please add a test that makes sure that this exception is thrown for some case where it needs to.

src/generic_datatype.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+  // Returns 0 if the constructor was not found
+  return num_selectors;
+}
+
+/*
+This function goes through every selector in the datatype and if
+finalized is set to false, it replaces the previously stored sort
+with new_sort
```
yoni206, line 153: great!

src/generic_solver.cpp
```diff
@@ -478,31 +481,128 @@ Sort GenericSolver::make_sort(SortKind sk, const SortVec & sorts) const
...
-  throw NotImplementedException("Generic Solvers do not support datatypes");
+  shared_ptr<GenericDatatypeDecl> gdt_decl =
+      static_pointer_cast<GenericDatatypeDecl>(d);
+  string dt_decl_name = gdt_decl->get_name();
+  assert(name_datatype_map->find(dt_decl_name) != name_datatype_map->end());
+  shared_ptr<GenericDatatype> curr_dt = (*name_datatype_map)[dt_decl_name];
+  if (name_sort_map->find(dt_decl_name) == name_sort_map->end())
+  {
+    // Exact functionality of make_genericsort, without the linking
+    // errors (undefined reference when I call a new make_generic_sort).
```
yoni206, line 492: huh? :) I don't understand this, but anyway this doesn't look like something that should get into `master`. Maybe it is time to fix this? If it doesn't work out, please let me know what the problem is offline.

src/generic_solver.cpp
```diff
@@ -779,8 +880,12 @@ Term GenericSolver::make_term(const Op op,
 
 Term GenericSolver::make_term(const Op op, const TermVec & terms) const
 {
+  cout << "this new fn called" << endl;
   Sort sort = compute_sort(op, this, terms);
+  cout << "computer sort" << endl;
   string repr = "(" + op.to_string();
+  cout << repr << endl;
+  cout << "tedt" << endl;
```
yoni206, line 888: please remove prints

src/generic_sort.cpp
```diff
@@ -14,9 +14,12 @@
 **/
 
 #include "generic_sort.h"
-#include <unordered_map>
+
 #include <functional>
+#include <unordered_map>
+
```
yoni206, line 20:
```suggestion
```

tests/test-generic-sort.cpp
```diff
@@ -70,5 +70,36 @@ int main()
...
+      make_shared<GenericDatatypeSort>(new2_dt);
+  assert(dt_sort != dt_sort2);
+  auto copy = dt_sort;
+  assert(dt_sort == copy);
+  cout << dt_sort->compute_string() << endl;
+  cout << dt_sort2->compute_string() << endl;
+  assert(dt_sort->to_string() != dt_sort2->to_string());
+  assert((dt_sort->get_sort_kind()) == (dt_sort2->get_sort_kind()));
+  assert(dt_sort->get_sort_kind() == DATATYPE);
+  assert(dt_sort2->get_sort_kind() == DATATYPE);
```
yoni206, line 102: Thanks for adding this test! Can you please add documentation comments for it as well? 
Like "create the datatype" before the block that creates it, "compare equal datatypes" before some assertions etc.

yoni206 (commented):
Great work, thank you! The new documentation helps a lot. Some more comments here.

include/generic_sort.h
afnaanhashmi: What about the compare() function for the generic_sorts? That still has a case where I need to compare sorts of sortkind DATATYPE. I struggled previously with trying to cast the object itself to access the generic datatype. Is doing so even possible?

include/generic_sort.h
afnaanhashmi: sort.h does not declare a compute_string function, and so generic_sort doesn't override anything for its compute_string() function. And so genericDatatypeSort doesn't have a 'base' virtual function to just override.

src/generic_datatype.cpp
afnaanhashmi: Great catch :)

include/generic_sort.h
yoni206: So this should be done not by casting actually, but similarly to here:
https://github.com/makaimann/smt-switch/blob/5c0c1a0f3ef1e4578f6a8dad066a155fa3eb15fd/include/generic_sort.h#L95

This function is declared for `GenericSort`, but by default throws an exception. Only in some of the inheritting classes, it is implemented, using fields that only these inheritting classes have.

Similarly, you can have a fuction called `getDatatype` or something like that. By default, it will throw an exception. But in `GenericDatatypeSort`, it will return the `Datatype` field. Then you can use this field to extract the name. Sounds good?

include/generic_sort.h
yoni206: I suggest we change that. Maybe add `= 0` in `GenericSort` for that funciton, to make sure that ehre is a default empty unction, and then mark it here is `override`.

include/generic_sort.h
afnaanhashmi: I'm still not following. How will I be able to use a function that's implemented for GenericDatatypeSort in GenericSort?

yoni206: You will define it for `GenericSort` and declare it with an implementation that throws an exception.
But on runtime, it will use the implementation of `GenericDatatypeSort`. 

This should have the exact structure of `get_uninterpreted_param_sorts`.

Is that clearer?

afnaanhashmi: Yes that does make sense. I'm surprised that it will use the GenericDatatypeSort on runtime. Why is that?

yoni206: It should always choose the implementation according to the actual type of the object. And I believe we will always `compare` a `DatatypeGenericSort` to a `DatatypeGenericSort`.

src/generic_sort.cpp
```diff
@@ -118,10 +120,19 @@ Sort make_generic_sort(SortKind sk, SortVec sorts)
   }
 }
 
+Sort make_generic_sort(Datatype & dt)
```
4tXJ7f, line 123:
```suggestion
Sort make_generic_sort(Datatype dt)
```

We need to remove the `&` to make `make_generic_sort` match the declaration in the header file. This will get rid of the undefined references in the linking step. With the `&`, we are effectively defining a different overload of the function `make_generic_sort` because `Datatype` and `Datatype&` are two different types. As a result, the linker cannot find the implementation of `Sort make_generic_sort(Datatype dt)`.

4tXJ7f (changes requested):

afnaanhashmi:
Closing. The update pr is #259

PR #252, opened 2021-06-26
Tseitin transformation

BhaskarGupta22:
Added code for tseitin_transformation


yoni206 (changes requested):
Hi, looking good, thanks! first round of comments.

src/utils.cpp
```diff
@@ -334,6 +334,186 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+
+Term to_cnf(Term formula, SmtSolver s, Sort boolsort){
+  srand(time(NULL));
+  std::map<Term, Term>ma;//map ma contains the mapping of each term to it's new symbol in tseytin's transformation
+  
+  //returning symbolic constants directly
+  if(formula->is_symbolic_const()){
+    return formula;
+  }
+  //This vector is going to be broken down into it's symbolic constants, as we need to know which symbols have already been used so that we don't create symbols with the same name
```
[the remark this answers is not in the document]
BhaskarGupta22: I had initially added a vector but had removed it, so removed this comment in the last commit

src/utils.cpp
```diff
@@ -334,6 +334,186 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
   }
 }
 
+
+
+
+
+Term term_gen(SmtSolver s, Sort boolsort, int& pt){
+  return (s->make_symbol("x"+std::to_string(pt++), boolsort));
```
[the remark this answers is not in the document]
BhaskarGupta22: I'll change it, for verifying the correctness it was easier to read the output with small string names.

src/utils.cpp
```diff
@@ -334,6 +334,186 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+  
+
+
+  static int pt=1;// a pointer which will decide the next symbols name
+  
+  TermVec vec={formula};//vec stores the formulas which yet need to be given a symbolic name
+  //reduced stores (c) <-> (a op b). (c) as the first term of the pair and (a op b) as the second term in the pair
+  std::vector<std::pair<Term, Term>>reduced;
+  bool fir=0;//If fir is true it mean's we have named the parent formula i.e the formula given to us
+  Term parent;//parent will contain the symbolic name of the parent formula
```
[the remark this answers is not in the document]
BhaskarGupta22: By symbolic name, I meant the new symbol which I generated to refer to the formula

src/utils.cpp
```diff
@@ -334,6 +334,186 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+  while(!vec.empty()){
```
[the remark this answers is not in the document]
BhaskarGupta22: No, I used the same idea as the earlier task(Of taking the last term, breaking it, and pushing the components back into the vector). Most of the code which I have written is regarding mapping to symbols, and for traversal, I have just written 5-6 lines. Should I use the walker for mapping?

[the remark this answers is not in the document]
yoni206: Yes I suggest you use the walker. Maybe it is just 5-6 lines now, but it might become more, and it is better to use an existing infrastructure rather than doing it again.

yoni206:
Thanks for the rewrite!
Could u please try to explain here why it is a problem to use a walker for this transformation?

src/utils.cpp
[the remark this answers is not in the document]
BhaskarGupta22: If t exists in the form c1 op c2, I need to map these formulas to new symbols, so x1 would now be x2 op x3. But now I have to add the formula x1<->(x2 op x3) in tseitin's transformation. I think if I use a walker I would need to traverse the term again to pair up x1 with (x2 op x3) as just mapping the subformulas with new symbols wouldn't be enough. So, in my code, I have simultaneously mapped the subformulas to new symbols and added the pair to a vector.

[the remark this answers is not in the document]
yoni206: I see. I am wondering whether @makaimann has some idea here. @makaimann , @BhaskarGupta22 is implementing Tseitin transformation. I am trying to push to a walker-based implementation but it is not clear whether walkers are a right fit for this task. If you have an idea of how to use a walker for this, please let us know. Otherwise, we will go forward without using a walker.

src/utils.cpp
```diff
@@ -332,8 +332,153 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
   }
+} 
+
+Term to_cnf(Term formula, SmtSolver s)
+{
+  
+  std::map<Term, Term>ma;//map ma contains the mapping of each term to it's new symbol in tseytin's transformation
+  
+  Sort boolsort = formula->get_sort();
+  assert(boolsort->get_sort_kind() == BOOL);
```
makaimann, line 343: A minor point here -- this won't always pass due to an edge case for some solvers. Some solvers alias booleans and bit-vectors of size one. And in smt-switch we (somewhat arbitrarily) picked to treat them as bit-vectors. In that case, even if it is a formula, the sort kind would actually be `BV` with width one. Using the logging solver handles this edge case (see [here](https://github.com/makaimann/smt-switch#logging-solvers) for more info), but since we don't know that we'll get a logging solver, I'd recommend the following:
```
Sort boolsort = s->make_sort(BOOL);
assert(formula->get_sort() == boolsort);
```
which should do what you want even if the underlying solver treats booleans and bit-vectors the same.

src/utils.cpp
```diff
@@ -332,8 +332,153 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+  
+  //reduced stores (c) <-> (a op b). (c) as the first term of the pair and (a op b) as the second term in the pair
+  std::vector<std::pair<Term, Term>>reduced;
+  
+  while(!vec.empty())
+  {
+    Term u = vec.back();
+    vec.pop_back();
+    smt::Op op = u->get_op();
+    if(op == smt::And || op == smt::Or || op == smt::Xor || op == smt::Implies)
```
makaimann, line 392: I could be missing if this is handled elsewhere, but you might have one other case: `Equal`. There's no `iff` in SMT-LIB, instead it's represented as an equality between booleans. E.g., `a <-> b` is actually `a = b`. So you might have to check if you have an equality between two booleans.

src/utils.cpp
```diff
@@ -332,8 +332,153 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+  }
+
+  static int pt = 1;// a pointer which will decide the next symbols name
+
+  auto term_gen = [&]
+  {
+    return (s->make_symbol("cnf__tseitin__fresh__var__" + std::to_string(pt++), boolsort));
+  };
+
+  auto give_symbolic_name=[&](Term t, bool push)
```
makaimann, line 360: I don't feel that strongly about this, but it might be a bit cleaner to have `give_symbolic_name` only handle names. For pushing onto the vector, you could either keep a separated `visited` cache, or just check the same conditions again separately. This is just something to consider.

src/utils.cpp
```diff
@@ -332,8 +332,153 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+      clauses.push_back(s->make_term(Or, s->make_term(Not, ri), fi));
+    }
+    else
+    {//((~a) v (~c)) and ((a) v (c))
+      Term le = (*(se->begin()));
+      clauses.push_back(s->make_term(Or, s->make_term(Not, le), s->make_term(Not, fi)));
+      clauses.push_back(s->make_term(Or, le, fi));  
+    }
+  }
+  //taking the and of all clauses generated to create the cnf
```
makaimann, line 468: This is fine, but the following should also work: `return  s->make_term(And, clauses);` If it doesn't, I can fix it :smile:

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,41 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
 
 
+
+TEST_P(UnitUtilDimacsTests, tseitin)
+{
+  Term p = s->make_symbol("p", boolsort);
+  Term q = s->make_symbol("q", boolsort);
+  Term r = s->make_symbol("r", boolsort);
+  Term t = s->make_symbol("t", boolsort);
+  if (s->get_solver_enum() == BTOR || s->get_solver_enum() == BZLA)
```
makaimann, line 239: These are the two solvers that alias booleans and bit-vectors of size one just FYI. Also, if incremental solving is an issue, you can enable it with: `s->set_opt("incremental", "true")`.

makaimann (commented):
Thanks @BhaskarGupta22 for all this work! It's looking really good. I left a few comments for you and @yoni206 to consider.

src/utils.cpp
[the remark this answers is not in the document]
makaimann: Sorry I didn't see this earlier. I might be missing something, but I think a walker would work fine for this. The walker does a pre- and post-order traversal. You would only be interested in the post-order traversal I believe. It would work something like this, create a new class that inherits from the `IdentityWalker`, and add whatever data structures you need (for example a mapping between terms and symbols). Then overload the virtual function `visit_term`. Within that function you would always update the cache if `preorder_` is false (i.e., it's in the post-order).

Then if you have something like `c1 op c2` that you want to name `x1`, you would first look up each of the arguments in the cache which would give you `x2 op x3`. Each of the children are guaranteed to have been processed already because you're in the post-order traversal. Then you could update the cache with `(c1 op c2) : x1`, and to a separate data structure save `x1 <-> (x2 op x3)`. 

Then you could query your walker object for those mappings (essentially your vector `reduced`) and create the clauses from there. All that being said, I'm not sure that it ends up being much less code because of the class boiler-plate. However, it automatically handles the DAG traversal, which you do manually by checking if it's already been mapped within `give_symbolic_name`.

Hopefully that's helpful. I think the walker might be cleaner, for similar reasons to my comment [here](https://github.com/makaimann/smt-switch/pull/252/files#r663207247)

tests/unit/unit-util.cpp
BhaskarGupta22: Thanks!

src/utils.cpp
BhaskarGupta22: It's working :). actually, I didn't know this was allowed and so now I binarised my formula before doing the tseitin transformation

src/utils.cpp
BhaskarGupta22: Thanks! I had missed this

src/utils.cpp
BhaskarGupta22: Changed it in the last commit

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,38 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+  Term p = s->make_symbol("p", boolsort);
+  Term q = s->make_symbol("q", boolsort);
+  Term r = s->make_symbol("r", boolsort);
+  Term t = s->make_symbol("t", boolsort);
+
+  //a=((p or q) and r) implies (not t)
+  Term a = s->make_term(Implies, s->make_term(And, s->make_term(Or, p, q), r), s->make_term(Not, t));
+  Term cnf1 = to_cnf(a, s);
+  s->assert_formula(cnf1);
+  Result r1 = s->check_sat();
```
yoni206, line 245: Here it would be wiser to do push, check-sat of the original formula, pop, and check-sat of the new formula, and then comparing the results.

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,38 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+  Result r2 = s->check_sat();
+  ASSERT_TRUE(r2.is_sat());
+  //c=((not p) and p)
+  Term c = s->make_term(And, s->make_term(Not, p), p);
+  Term cnf3 = to_cnf(c, s);
+  s->assert_formula(cnf3);
+  Result r3 = s->check_sat();
+  ASSERT_FALSE(r3.is_sat());
+
+
```
yoni206, line 260: It would be great to check some general smt-lib files on this and verify that we get the right result. 
There are smtlib tests under here: https://github.com/makaimann/smt-switch/tree/master/tests/smt2

Can you look into how these files are tested and maybe come up with some less trivial tests? Then we can just check that the results are the expected ones.

yoni206 (changes requested):
Thanks for all the changes!
Looks good!
So far, I skipped the actual transformation in `to_cnf`, but the walkers look good. I'll review the final part after the current comments are addressed.

src/utils.cpp
```diff
@@ -334,6 +335,203 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
   }
 }
 
+//mapping each subformula with a new name and returning a vector of pair of terms. 
+//Each pair consists of the parent term and it's children(for each subformula)
+class traversal : public IdentityWalker
+{
+ public:
+  std::vector<std::pair<Term, Term>> reduced;
```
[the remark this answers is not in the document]
BhaskarGupta22: I am accessing this field from to_cnf function, so shouldn't I keep it public?

src/utils.cpp
```diff
@@ -334,6 +335,203 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+                     // the children to generate the new term
+    {
+      smt::Op op = term->get_op();
+      if (!op.is_null())
+      {
+          std::vector<Term>vec;//a vector of all children
+          for(auto u:term)
+          {
+            Term term_name;
+            query_cache(u, term_name);//finding the new name of each child from the cache
```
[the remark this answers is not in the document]
BhaskarGupta22: If it is surely present in the cache, should I still assert it?

src/utils.cpp
```diff
@@ -334,6 +335,203 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+            vec.push_back(term_name);
+          }
+          Term term_name;
+          bool present=query_cache(term, term_name);//true if we found the name of the term in the cached
```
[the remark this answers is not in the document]
BhaskarGupta22: I don't think so, I'll change it

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,38 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+  s->assert_formula(cnf2);
+  Result r2 = s->check_sat();
+  ASSERT_TRUE(r2.is_sat());
+  //c=((not p) and p)
+  Term c = s->make_term(And, s->make_term(Not, p), p);
+  Term cnf3 = to_cnf(c, s);
+  s->assert_formula(cnf3);
+  Result r3 = s->check_sat();
+  ASSERT_FALSE(r3.is_sat());
+
```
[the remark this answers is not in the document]
BhaskarGupta22: The actual cnf had become too long, so I thought it would be unreadable. I had checked that it was exactly what I had expected. I'll add it as a test though.

src/utils.cpp
```diff
@@ -334,6 +335,203 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
   }
 }
 
+//mapping each subformula with a new name and returning a vector of pair of terms. 
+//Each pair consists of the parent term and it's children(for each subformula)
+class traversal : public IdentityWalker
+{
+ public:
+  std::vector<std::pair<Term, Term>> reduced;
+  static int pt;  // the static variable to always produce distinct symbols
```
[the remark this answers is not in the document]
yoni206: @makaimann an idea we might have here is to expose a unique pointer to the `traversal` class. Would that solve this issue? And do you think this would be better than a static member?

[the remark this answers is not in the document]
makaimann: Right, so it seems like you would want a "static" member for each solver (e.g., if you had two `traversal` objects using the same solver). But otherwise, you don't want a static member because you can reuse symbol names. It's not the most performance efficient, but for a first pass, I think a non-static `pt` with a `get_unique_symbol` function would be fine. Something like this (didn't check if it would compile):

```
Term traversal::get_unique_symbol()
{
  while(true)
  {
     try
     {
        return solver_->make_symbol("fresh_symbol_" + to_string(pt), boolsort_);
     }
     catch(IncorrectUsageException & e)
     {
       ++pt;
     }
  }
}
```

I don't think you need to worry about the edge case (enumerating all possible pt values). But you could always put a condition in the while-loop for that if needed. Also, I'd prefer the class name to be capitalized: `Traversal`. And maybe a more descriptive name would be good as well.

src/utils.cpp
```diff
@@ -334,6 +335,203 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+    }
+
+    return Walker_Continue;
+  }
+  Term acc_cache(Term term)
+  {
+    Term ne;
+    query_cache(term, ne);
+    return ne;
+  }
```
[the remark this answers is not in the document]
BhaskarGupta22: I have used it in to_cnf function

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
 }
 
 
+
+TEST_P(UnitUtilDimacsTests, tseitin)
```
yoni206, line 233: Can you clang-format this file?

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+  Term p = s->make_symbol("p", boolsort);
+  Term q = s->make_symbol("q", boolsort);
+  Term r = s->make_symbol("r", boolsort);
+  Term t = s->make_symbol("t", boolsort);
+  
+  //a=((p or q) and r) implies (not t)
+  Term a = s->make_term(Implies, s->make_term(And, s->make_term(Or, p, q), r), s->make_term(Not, t));
+  Term cnf1 = to_cnf(a, s);
+  string st1 = cnf1->to_string();
+  string ans1 =
```
yoni206, line 245: It would be much better to compare terms and not strings, in case we change something in the string representation at some point. Could you construct a `Term` object that represents the expected result and compare it to the actual result (and not to its string representation)?

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+      "cnf_formula_new_2 (not cnf_formula_new_3)) (or r (not "
+      "cnf_formula_new_3))) (or (or (not cnf_formula_new_3) cnf_formula_new_1) "
+      "(not cnf_formula_new_4)) (or cnf_formula_new_3 cnf_formula_new_4) (or "
+      "(not cnf_formula_new_1) cnf_formula_new_4)))";
+  ASSERT_TRUE(ans1 == st1)<<s->get_solver_enum();
+  //b=(not (p xor q))
+  Term b = s->make_term(Not, s->make_term(Xor, p, q));
+  Term cnf2 = to_cnf(b, s);
+  string st2 = cnf2->to_string();
+  string ans2 =
```
yoni206, line 259: ditto.

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+      "cnf_formula_new_5)) (or (or p q) (not cnf_formula_new_5)) (or (or "
+      "cnf_formula_new_5 q) (not p)) (or (or cnf_formula_new_5 p) (not q)) (or "
+      "(not cnf_formula_new_5) (not cnf_formula_new_6)) (or cnf_formula_new_5 "
+      "cnf_formula_new_6)))";
+  ASSERT_TRUE(ans2 == st2);
+  //c=((not p) and p)
+  Term c = s->make_term(And, s->make_term(Not, p), p);
+  Term cnf3 = to_cnf(c, s);
+  string st3 = cnf3->to_string();
+  string ans3 =
```
yoni206, line 270: ditto.

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+      "(and cnf_formula_new_8 (and (or (not p) (not cnf_formula_new_7)) (or p "
+      "cnf_formula_new_7) (or cnf_formula_new_8 (not cnf_formula_new_7) (not "
+      "p)) (and (or cnf_formula_new_7 (not cnf_formula_new_8)) (or p (not "
+      "cnf_formula_new_8)))))";
+  ASSERT_TRUE(ans3 == st3)<<ans3<<endl<<endl<<s->get_solver_enum();
+}
+
```
yoni206, line 277: In addition to checking that the transformation does what it should do, let's also add an assertion that makes sure that the solver answers the same about the original formula and the transformed formula.
For example, `s->assert_formula(s); s->check_sat()` and `s->assert_formula(cnf3); s->check_sat()` should either both be SAT or both be UNSAT.

Please add these checks for each one of the formulas. This will test the encoding itself, while the term comparison will test its implementation. Please let me know if something is not clear.

tests/unit/unit-util.cpp
```diff
@@ -229,6 +229,53 @@ TEST_P(UnitUtilDimacsTests, cnf_to_dimacs)
...
+
```
yoni206, line 278: Let's add some more simple tests and corner cases as inputs:
1. An input that is already in cnf
2. the constants `true/false`
3. A literal
4. A clause
5. A formula that contains something that is not a Boolean variable (then make sure that an exception is throen).

yoni206 (changes requested):
Thanks for updating!
I think that the implementation is fine, and I've added comments about the tests.

tests/unit/unit-util.cpp
BhaskarGupta22: Ok sure.
1. The formula which I would output would be different, is that fine? I few extra variables will be added.
2. Do you mean a large formula contains within itself true/false. As of now, I have not handled this case but I can implement it.

tests/unit/unit-util.cpp
BhaskarGupta22: I will not be able to create the same symbols using the same solver(As an exception would be thrown). Should I use 2 solvers?

tests/unit/unit-util.cpp
yoni206: Well, as an optimization, I suggest you add a check before transforming, and if the input is already in CNF, you can just return the input itself. 

I suggest handling true/false, they do occur in practical encodings, so supporting it is desirable. With CNF, they can be simulated, e.g., by adding a unit clause with a single fresh variable and another unit clause with a single negated fresh variable, and these can represent the constants.

And for the test, I recommend testing both the simple formulas `true` and `false` as well as their incorporation in other formulas.

BhaskarGupta22: Ok, could you also check out my reply to the comment regarding asserting strings or terms?

tests/unit/unit-util.cpp
BhaskarGupta22: Will add this too

tests/unit/unit-util.cpp
yoni206: Yes, good point!

yoni206: But actually, I think that the terms will not be equal if they are created by two different solvers, but they should be based on the same solver, that is, two instances of cvc4, or two instances of z3, etc.

tests/unit/unit-util.cpp
yoni206: Yes, I did!

tests/unit/unit-util.cpp
BhaskarGupta22: Ok!

BhaskarGupta22: Should I create 2 solvers s1, s2(of the same solver) in the class of the test, or is that not allowed?

yoni206:
Yes, that what I suggest to try.

src/utils.cpp
```diff
@@ -334,6 +335,621 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+        for (auto u : term)
+        {
+          Term term_name;
+          bool present = query_cache(
+              u,
+              term_name);  // finding the new name of each child from the cache
+          assert(present == true);
+          vec.push_back(term_name);
+        }
+        Term term_name;
```
yoni206, line 385: It is a bit confusing to have again this declaration. If it is conceptually different, I suggest using a different name.

src/utils.cpp
```diff
@@ -334,6 +335,621 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+{
+ public:
+  EliminateBooleanConstants(SmtSolver solver_)
+      : IdentityWalker{ solver_, false }
+  {
+  }
+  WalkerStepResult visit_term(Term & term)
+  {
+    auto is_true = [&](Term t) {  // If the term is "true"
+      return ((t->to_string()) == "true");
```
yoni206, line 465: It is better not to use string comparisons but term comparisons. You can create beforehand a `true` term and a `false` term, and compare to them. Maybe some day we decide to print this as `TRUE` or something like that, and then this code will break. The value of `t` should not depend on its string representation.

src/utils.cpp
```diff
@@ -334,6 +335,621 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+    };
+    auto is_false = [&](Term t) {  // If the term is "false"
+      return ((t->to_string()) == "false");
```
yoni206, line 468: ditto.

src/utils.cpp
```diff
@@ -334,6 +335,621 @@ void cnf_to_dimacs(Term cnf, std::ostringstream & y)
...
+                                   // as we are doing a preorder traversal
+
+          if (is_true(term_name))
+          {  // not(false)=true
+            save_in_cache(term, fa);
+          }
+          else if (is_false(term_name))
+          {  // not(true)=false
+            save_in_cache(term, tr);
+          }
```
yoni206, line 492: It is unclear how `true` and `false` are eliminated here -- they are stored in the cache. Can you clarify inside a comment?

yoni206 (commented):
Thanks for all of these. Some new comments on the new code.

src/utils.cpp
```diff
@@ -452,6 +450,11 @@ class XorBinarize : public IdentityWalker
 
 // Given a boolean formula, removes terms "true" and "false" to give a formula
 // without adding new symbols
+
+// The way true and false are eliminated is by doing a preorder traversal. When
+// I am on a certain node, The children are already reducecd. Reduced means that
+// they are a term without true or false, or just true, or just false.
```
yoni206, line 456: and what do you do with just true and just false terms?

BhaskarGupta22: If the entire formula is reduced to true or false, I return it. Should I do something else?

yoni206: But then how to you print the dimacs?

BhaskarGupta22: If it is reduced to true/false. I will not pass it to print dimacs

BhaskarGupta22: Sorry, I have handled the case of the term being true/false in the function cnf_to_dimacs

BhaskarGupta22: True will be printed as 
: p cnf 0 0
False as:
p cnf 0 1
0

yoni206:
got it!

yoni206 (approved):
Great, thanks!

PR #254, opened 2021-07-19
Ipasir backend

BhaskarGupta22:
Added ipasir-setup, made changes to configure.sh and CMakeLists.txt, so that the first 2 commands ./contrib/setup-ipasir.sh and ./configure.sh --ipasir can  run smoothly. Ipasir directory is still incomplete


CyanoKobalamyne:
Ipasir seems to be a SAT solving interface. As we currently only support SMT solvers and have no immediate plans on adding support for SAT, I will close this PR. Please let us know if this is a mistake.

PR #255, opened 2021-07-20
Fix: keep track of symbol declarations in Yices2 and Z3

makaimann:
To match other solvers, keep track of declared symbols and throw exception if trying to redeclare. I looked into the APIs a bit and it doesn't seem like there's a way to do this without externally keeping a set of used symbols. For Yices2, this is because it has a global symbol table and we want to only enforce this for symbols of a given solver instance. For Z3, it has a method for creating a fresh constant, but I couldn't see a method to directly query the symbol table to see if a symbol has been used before.


yoni206 (commented):
Thanks for this!
Will it work for incremental? I think that this input is legal, but would trigger an exception now:
https://cvc4.github.io/app/#temp_8027ba2b-922b-4f47-a8ae-4e30db39f5b4

no?

If yes, isn't this a problem? If not, maybe add a corresponding test?

makaimann:
Hmm that is a good point. Currently the other solver backends would throw an exception there. I think this might merit a longer discussion, but my inclination is to merge this so that all the solvers behave the same and consider changing it later. Certainly the safest thing to do is be fully compliant with SMT-LIB. However, I'm not sure it makes as much sense in the API. If you wanted to use `x` again, just keep a reference to the original `x`. Although, I could see an argument for using the solver with some local context and not worrying about keeping track of all the declared symbols...

The main issue is ensuring that all solvers can support that. But if they all support SMT-LIB, then in theory that should be no problem.

yoni206:
Oh really the other solvers will throw an exception? Even cvc5? how come?

I agree with the point about the API but we also have a parser now :)

I am also in favor of this solution, but wonder whether / where this point should be documented.

makaimann:
Other solver `backends`, so they only behave that way because I wrote them that way haha. `cvc5` allows re-declaring symbols even without a context level change.

The parser already handles that to some degree (it currently doesn't allow symbols of different sorts to have the same name, but otherwise aligns with SMT-LIB). I suppose the question is whether that should be handled at the solver level or parser level. If we do it at the solver level it's a little bit of extra data / infrastructure even if it's not being used with the parser.

yoni206 (approved):
Thanks for this and for the answers!

makaimann:
Thanks for the review and good questions! I'll make an issue and we can revisit it later.

PR #257, opened 2021-07-22
Enable non-zero arity uninterpreted sorts in CVC4 backend

makaimann:


yoni206 (approved):
Great!
Maybe consider adding a test with arity -1, to make sure we get an exception.

makaimann:
Thanks for the review! Good idea, I added a test and fix with 7119e78aee3bdee085424c42bc92e22cbe570222. I was expecting CVC4 to complain if not provided enough parameters, but it doesn't so I just added the check myself.

PR #259, opened 2021-07-22
Datatype Sorts + Related Functionality

afnaanhashmi:


src/generic_solver.cpp
```diff
@@ -495,8 +491,8 @@ Sort GenericSolver::make_sort(const DatatypeDecl & d) const
     // Exact functionality of make_genericsort, without the linking
     // errors (undefined reference when I call a new
     // make_generic_sort).
-    //Sort dt_sort = make_generic_sort(curr_dt);
-    Sort dt_sort = make_shared<GenericDatatypeSort>(curr_dt);
+    Sort dt_sort = make_generic_sort(curr_dt);
+    //Sort dt_sort = make_shared<GenericDatatypeSort>(curr_dt);
```
yoni206: ...

yoni206 (approved):
Looks good. Found some commented out code, please go over and check there is no other isntances of this.
Other than that looks good, but lets wait for the CI to finish running before we merge.

src/generic_solver.cpp
afnaanhashmi: I am confused by this, because looking through my branch and the most recent push, this commented line is nowhere to be found.

yoni206:
Is it the correct (new) branch? And did u push to the right remote branch?

afnaanhashmi:
> Is it the correct (new) branch? And did u push to the right remote branch?
> […](#)
I think that was an older, outdated commit. I'm pretty sure the two most up-to-date pushes should not have that as a line.

PR #266, opened 2021-08-04
Add symbol table to solver backends

makaimann:
This PR adds a `get_symbol` feature to `AbsSmtSolver`. This allows a user to look up a symbol by name. This can be very useful for term translation as well, so we can look up symbols instead of keeping track of the mapping via externally populated caches (in the case where the target solver has already been used).

Note, this implementation lets each solver backend deal with how to implement `get_symbol`. It results in some code duplication, but I think it's the easiest to understand. Especially because each backend solver has to own the symbol table. If the symbol table were stored in `AbsSmtSolver` then it would get destructed _after_ the backend solver which has bad refcounting implications for many solvers.

### DIFF
--- a/include/solver.h
+++ b/include/solver.h
@@ -231,7 +231,17 @@ class AbsSmtSolver
 
   /** Look up a symbol by name.
    *  If no symbol of that name has been declared, throws
-   * IncorrectUsageException
+   *  IncorrectUsageException
+   *
+   *  Allows a user to look up a symbol by name. This can be very useful for term
+   *  translation, since we can look up symbols instead of keeping track of the
+   *  mapping via externally populated caches (in the case where the target
+   *  solver has already been used).
+   *
+   *  Note, solver backend deals with the implementation. The main motivation for
+   *  this is that each backend solver has to own the symbol table. If the symbol
+   *  table were stored in AbsSmtSolver then it would get destructed after the
+   *  backend solver which has bad refcounting implications for many solvers.
    *
    *  @param name the name of the symbol to look up
    *  @return the Term representation of the corresponding symbol
--- a/include/solver.h
+++ b/include/solver.h
@@ -229,6 +229,15 @@ class AbsSmtSolver
    */
   virtual Term make_symbol(const std::string name, const Sort & sort) = 0;
 
+  /** Look up a symbol by name.
+   *  If no symbol of that name has been declared, throws
+   * IncorrectUsageException
+   *
+   *  @param name the name of the symbol to look up
+   *  @return the Term representation of the corresponding symbol
+   */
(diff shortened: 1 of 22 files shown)


include/solver.h
```diff
@@ -229,6 +229,15 @@ class AbsSmtSolver
    */
   virtual Term make_symbol(const std::string name, const Sort & sort) = 0;
 
+  /** Look up a symbol by name.
+   *  If no symbol of that name has been declared, throws
+   * IncorrectUsageException
+   *
+   *  @param name the name of the symbol to look up
+   *  @return the Term representation of the corresponding symbol
+   */
```
yoni206, line 238: I recommend expanding here, maybe with some of the material in the PR description.

src/generic_solver.cpp
```diff
@@ -817,28 +817,45 @@ Term GenericSolver::make_term(const Term & val, const Sort & sort) const
 
 Term GenericSolver::make_symbol(const string name, const Sort & sort)
 {
+  // always put pipes around name in case there
+  // are special symbols / spaces
+  string piped_name = "|" + name + "|";
```
yoni206, line 822: nice :)

tests/unit/unit-symbol.cpp
```diff
@@ -56,6 +56,27 @@ TEST_P(UnitSymbolTests, RedeclareException)
...
+  EXPECT_EQ(a, s->get_symbol("a"));
+
+  string funky_name = "strange @ name!";
+  string funky_name_fun = "strange Fun _ name$";
+  Term funky_sym = s->make_symbol(funky_name, boolsort);
+  Term funky_sym_fun = s->make_symbol(funky_name_fun, funsort);
+
+  EXPECT_EQ(funky_sym, s->get_symbol(funky_name));
+  EXPECT_EQ(funky_sym_fun, s->get_symbol(funky_name_fun));
+}
```
yoni206, line 78: Maybe also a test that verifies that exceptions are thrown when needed?

yoni206 (approved):
Looks good, minor suggestion inline.

include/solver.h
makaimann: good idea! added with b4df4d3cf1a07327a7b52b3cf5ce98494ebd3223

src/generic_solver.cpp
makaimann: :)

tests/unit/unit-symbol.cpp
makaimann: Good idea! Added with b00cde7162139a73aebaedb3432efc6307e868c5.

makaimann:
Thanks for the review! I've addressed your feedback with the last two commits. I'll wait for GitHub Actions to finish and then merge it.

PR #271, opened 2021-08-05
Mutually recursive datatype support in CVC4 backend

makaimann:
This PR adds support for mutually recursive datatypes in the CVC4 backend. This is accomplished by using uninterpreted sorts as unresolved datatype sorts to allow referencing them before completely defining the datatype. This is a direct wrapping of the CVC4 API's approach.

### DIFF
--- a/tests/test-dt.cpp
+++ b/tests/test-dt.cpp
@@ -41,6 +41,43 @@ class DTTests : public ::testing::Test,
   Sort intsort;
 };
 
+TEST_P(DTTests, DeclareSimpleList)
+{
+  SolverConfiguration sc = GetParam();
+  if (sc.is_logging_solver || s->get_solver_enum() == GENERIC_SOLVER)
+  {
+    return;
+  }
+
+  DatatypeDecl listSpec = s->make_datatype_decl("list");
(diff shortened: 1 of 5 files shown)


tests/test-dt.cpp
```diff
@@ -41,6 +41,43 @@ class DTTests : public ::testing::Test,
...
+TEST_P(DTTests, DeclareSimpleListWithForwardRef)
+{
+  SolverConfiguration sc = GetParam();
+  if (sc.is_logging_solver || s->get_solver_enum() == GENERIC_SOLVER)
+  {
+    return;
+  }
+
+  DatatypeDecl listSpec = s->make_datatype_decl("list");
+  Sort forward_ref_listsort = s->make_sort("list", 0);
```
4tXJ7f, line 71: Maybe not necessary for this PR but what do you think about adding a `make_forward_ref` convenience function that takes a `DatatypeDecl` as an argument and does essentially this? The advantage would be that the user can't get the names wrong.

4tXJ7f (approved):
LGTM

tests/test-dt.cpp
makaimann: That's a really good idea! It seems more "future-proof" as well since not every solver necessarily uses uninterpreted sorts for this purpose.

PR #276, opened 2021-08-23
Add getter for context level

makaimann:


4tXJ7f (approved):
LGTM. An alternative would be to implement the functionality at the level of solver.h and then to override it as needed for the individual solvers.

makaimann:
Thanks for the review! That's true. I've been thinking about moving in that direction to reduce effort. One general issue I have is that you often want to call some base version of the function, but I worry about making it complicated to remember what to call. In this case, I mean if you rely on a base implementation, it has to update `context_level` in `push` and `pop` as well. So when those functions are overridden you could get a mismatch.

Do you know of some design pattern that would treat functions like constructors? I.e. calls the base function first and then the virtual override after. I'm not sure if that is a good idea even if it is possible though.

PR #278, opened 2021-10-11
Explicitly find PythonLibs

makaimann:
There was a case on Mac where the correct python interpreter was found, but the `scikit-build` CMake files pulled in a `python2.7` library. This PR explicitly looks for the Python library with the same version, separately from the `scikit-build` infrastructure.


lonsing (approved):
Looks good to me.

makaimann:
Thanks for the review!

PR #287, opened 2021-11-11
Support time-limit in several backends

makaimann:
This PR adds support for a time-limit option (in seconds) for `bitwuzla`, `cvc5`, `yices2`, and `z3`. Support in other backends can be added in a future PR.


makaimann:
Hey @mpreiner, I was wondering if you'd mind reviewing this PR? I'm not sure if I approached it in a remotely sane way so I'd appreciate your feedback on the time-limit support in the `bitwuzla` backend.

Additionally, I'm using the `tlimit-per` option in the `cvc5` backend, but it only passes the test if I set it to a very small timeout. My guess is that it's spending a lot of time in preprocessing and doesn't check for the timeout. I was curious if you have any thoughts on that? I can look into it more and/or post an issue on `cvc5` if you think that's warranted.

PR #288, opened 2021-11-11
Support get-unsat-assumptions in Z3 backend

makaimann:
This PR adds support and tests `get_unsat_assumptions` for the Z3 backend.

### DIFF
--- a/msat/include/msat_solver.h
+++ b/msat/include/msat_solver.h
@@ -169,9 +169,12 @@ class MsatSolver : public AbsSmtSolver
   size_t max_assump_clauses_;  ///< number of assumption clauses before clearing
                                ///< them
   bool last_query_assuming;    ///< set to true if last query was
-                             ///< check-sat-assuming used to respect
-                             ///< get_unsat_assumptions interface (will complain
-                             ///< if not called after check-sat-assuming)
+                               ///< check-sat-assuming (as opposed to
+                               ///< just check-sat).
+                               ///< This boolean is used to respect the
+                               ///< get_unsat_assumptions interface (will
+                               ///< complain if not called after
+                               ///< check-sat-assuming).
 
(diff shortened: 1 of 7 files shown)


msat/include/msat_solver.h
```diff
@@ -169,6 +168,10 @@ class MsatSolver : public AbsSmtSolver
   size_t num_assump_clauses_;  ///< counts how many assumption clauses are added
   size_t max_assump_clauses_;  ///< number of assumption clauses before clearing
                                ///< them
+  bool last_query_assuming;    ///< set to true if last query was
+                             ///< check-sat-assuming used to respect
+                             ///< get_unsat_assumptions interface (will complain
+                             ///< if not called after check-sat-assuming)
```
yoni206, line 174: I don’t really understand the comments here. Could you try to clarify?

tests/python/test_unsat_core.py
```diff
@@ -18,7 +18,7 @@
 import smt_switch as ss
 
 # z3 backend doesn't support unsat assumptions yet
```
yoni206, line 20: Can this be deleted?

yoni206 (approved):
LGTM, minor comments,

msat/include/msat_solver.h
makaimann: Oops, this is partially because of formatting. I keep forgetting to check this. It makes slightly more sense with new lines, but clang-format seems to join them. I improved the wording and spacing with fabffb677ced3bb0aca1130e3bbbb2a2b41d889f.

tests/python/test_unsat_core.py
makaimann: Yep, good catch! Fixed with b49f50d01f6a3b5616e8f6cb23c390d45fddab2b.

PR #295, opened 2022-01-06
Build CVC5 using $(nproc) threads

georgerennie:
This works on Darwin or Linux in line with setup-bitwuzla.sh and setup-btor.sh

### DIFF
--- a/contrib/setup-cvc5.sh
+++ b/contrib/setup-cvc5.sh
@@ -7,6 +7,14 @@ mkdir -p $DEPS
 
 CVC5_VERSION=77d0bec48a745e3c4acd65085f9c59bdfceed6c0
 
+if [ "$(uname)" == "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


yoni206:
Thanks for this PR and sorry for the late reply. It looks good to me and I tested it locally. Once CI passes I'll approve and merge. @makaimann FYI.

yoni206:
@makaimann checks fail here — is this related to the issue on CI that you are fixing?

makaimann:
https://github.com/stanford-centaur/smt-switch/pull/298 made it so cvc5 is built in CI. If you merge master into this branch, it should pass. Would be good to check though, I’ve had issues with parallel builds in CI before.

Thanks for the PR!

yoni206:
@georgerennie -- would it be possible for you to merge current master into your branch and push, so that the PR is synced with current master? This will allow us to check CI (which is now fixed on master).

yoni206 (approved):

PR #296, opened 2022-01-06
Add BUILD_TESTS option to CMakeLists.txt

georgerennie:
Defaults to ON, when disabled prevents googletest being downloaded and tests being compiled.
When using smt-switch as a library there is generally no need to download googletest and run its test suite.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,6 @@ if (BUILD_BITWUZLA OR BUILD_CVC5 OR BUILD_MSAT OR BUILD_YICES2 OR BUILD_Z3)
   find_package(GMP REQUIRED)
 endif()
 
-# build testing infrastructure
-enable_testing()
-
 # should we build boolector?
 option (BUILD_BTOR
   "Should we build the libraries for boolector" OFF)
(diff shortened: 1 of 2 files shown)


georgerennie:
So by default the tests will be built, as before, but to disable tests being built you can call cmake with `-DBUILD_TESTS=OFF`. When using `configure.sh`, it automatically passes -D options through to cmake, so you can just call `./configure.sh -DBUILD_TESTS=OFF` (with the relevant other `configure.sh` flags).

I can add an extra flag to `configure.sh` (such as `--without-tests`) to pass this setting through to cmake in the same way the other settings are passed through if that is something that would be useful to do.

yoni206:
Thanks for your response. I think it would be a good idea to set this through a `configure.sh` flag.

georgerennie:
I've added a --without-tests flag to conigure.sh that passes through the setting to cmake so the tests arent built. This builds as expected locally for me but I imagine it will fail in CI until the other issue with the CVC5 CI build is fixed

yoni206:
@georgerennie -- would it be possible for you to merge current master into your branch and push, so that the PR is synced with current master? This will allow us to check CI (which is now fixed on master).

georgerennie:
I have rebased and pushed this PR and #295

yoni206:
Great, thanks! Waiting for CI to finish on both.

yoni206 (approved):

PR #297, opened 2022-01-23
Require a list or set in check-sat-assuming through Python interface

makaimann:
It can be easy to accidentally pass a single term to check_sat_assuming and not get an error. There was no error because terms are iterable, so it adds all the children as assumptions. However, this is typically not the desired behavior and can be very confusing.

### DIFF
--- a/python/smt_switch/smt_switch_imp.pxi
+++ b/python/smt_switch/smt_switch_imp.pxi
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from collections.abc import Iterable, Sequence
 
 from cython.operator cimport dereference as dref
 from libcpp.string cimport string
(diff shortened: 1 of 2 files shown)


yoni206 (approved):
Great, thanks for this!

makaimann:
Thanks for the review! The cvc5 tests are failing because I lost my file hosting where it downloaded pre-compiled versions from. I’ll look into fixing this and then will merge.

yoni206:
Cool.
What do you think about blocking the ability to merge when checks are unsuccessful?

makaimann:
Sorry for the slow response! Good idea. I think that branch protection is already on, but it looks like you can still override it. Let me know if there's something else I should change.

yoni206:
Is this turned on?

https://stackoverflow.com/a/58655352/1364765

PR #301, opened 2022-06-29
:bug: deps path fixed

Moorvan:
When building static library of smt-switch with bitwuzla, it didn't work with this version because of this little bug.


yoni206 (commented):
LGTM, and thanks for this!
I am a bit hesitant to promote to a newer commits of solvers, but, if CI passes then I think it is fine. @makaimann do you agree?

yoni206:
@Moorvan -- would it be possible for you to merge current master into your branch and push, so that the PR is synced with current master? This will allow us to check CI (which is now fixed on master).

makaimann:
> I am a bit hesitant to promote to a newer commits of solvers, but, if CI passes then I think it is fine. @makaimann do you agree?

Sorry for the slow response. I agree that as long as CI passes then it's good to go. There is risk of introducing issues of course, but it's also best to keep up to date. If it's not caught by CI, then we can add a new tests when/if issues show up.

Moorvan:
Okay, I've done it.

> @Moorvan -- would it be possible for you to merge current master into your branch and push, so that the PR is synced with current master? This will allow us to check CI (which is now fixed on master).

yoni206 (approved):

PR #309, opened 2022-10-25
print 1-bit bit-vectors correctly when using bitwuzla

### LINKED ISSUE #308: Calling `to_int` on a bitvector of size 1 causes a crash
It seems like `to_int` assumes that the bitvector gets converted to a string of the form `#bxxx`, but for bitvectors of size 1 it is actually `true` or `false` (at least when using the bitwuzla solver). For example, the following code causes a crash in `to_int`:

```cpp
SmtSolver solver = BitwuzlaSolverFactory::create(false);
solver->set_opt("produce-models", "true");
solver->set_opt("incremental", "true");

Sort sort = solver->make_sort(BV, 1);
Term x = solver->make_symbol("x", sort);
solver->check_sat();
uint64_t i = solver->get_value(x)->to_int();
std::cout << i << "\n";
```

I would instead expect `to_int` to return `0` or `1`.

yoni206:
Bitwuzla prints the bit-vectors `0` and `1` as `false` and `true`, respectively.
This PR wraps the result we get from bitwuzla by printing `0` and `1` in these cases.
fixes #308 and includes the benchmark provided there.
@barrettcw


makaimann (approved):
Looks great to me! Thanks for the fix. I see that CI is failing due to the unrelated issue we’ve discussed offline. I’m marking this approved so it can be merged as soon as that is fixed, or even before, at your discretion.

tests/test-bv.cpp
```diff
@@ -0,0 +1,64 @@
...
+TEST_P(BVTests, to_int)
+{
+  Sort sort1 = s->make_sort(BV, 1);
+  Sort sort2 = s->make_sort(BV, 2);
+  Term x1 = s->make_symbol("x1", sort1);
+  Term x2 = s->make_symbol("x2", sort2);
+  s->check_sat();
+  uint64_t i1 = s->get_value(x1)->to_int();
+  uint64_t i2 = s->get_value(x2)->to_int();
+  std::cout << i1 << "\n";
```
ahmed-irfan, line 52: do we need to print them in tests?

ahmed-irfan (approved):
One minor comment... 
Overall LGTM.

tests/test-bv.cpp
yoni206: Thanks for the review! Addressed 
074d306.

yoni206:
closing and opening a new PR.

PR #311, opened 2022-11-02
Install older version of setuptools on macOS CI

makaimann:
Python bindings installation is failing on macOS, likely due to an updated `setuptools`. Install an older version to temporarily avoid this issue.

See https://github.com/upscale-project/pono/pull/302 and https://github.com/pypa/setuptools/issues/3227 for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,13 @@ jobs:
             gmp \
             gperf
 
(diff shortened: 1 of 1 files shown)


makaimann:
Closing for now, because https://github.com/stanford-centaur/smt-switch/pull/312 is hopefully a better fix.

PR #312, opened 2022-11-02
Specify packages and py_modules for updated setuptools

makaimann:
The macOS CI build is failing to `pip` install the Python bindings due to a breaking change in the latest `setuptools` version. See https://setuptools.pypa.io/en/latest/history.html#v61-0-0 for more details. The fix is to explicitly specify the `packages` and `py_modules` keywords in the `setup.py`.


makaimann:
It looks like this PR successfully addressed the `setup tools 61` issue. Thanks @lonsing for the pointer to that issue. Unfortunately, I'm getting other CI failures that don't make sense to me.

One time there was a C++ test that failed (disjoint-set), another time it couldn't find `Cython` but only on the macOS btor run, not the other macOS runs. Very bizarre. It might make sense to merge this even if it's failing and address the other issues in a different PR? Or people can push other fixes to this branch. What do you think @yoni206?

yoni206:
Thanks @makaimann !
To me it seems like your fix enables us to find other problems, that indeed seem unrelated.
In particular, failing a regression is already progress, as before we couldn't pass the installation faze. 
I suggest we merge, and then deal with the new problems in separate PRs.

yoni206 (approved):

makaimann:
Great, sounds good. I'll merge now. Is there anyone who can take a look at the failing regression test? I can't reproduce it locally (same configuration arguments on my Mac laptop). Very strange.

yoni206:
The Cython thing is also happening in the cvc5 ci. I'll see how it is fixed there and will try to follow.

yoni206:
For the failing regression - I'll try to reproduce locally. Can you please specify the configuration? I can only see the Cython failure now on gh actions.

yoni206:
Can't reproduce the failing regression locally.

PR #315, opened 2022-11-18
update disjoint test 

ahmed-irfan:
This PR updates the disjointset test so that it works with different compiler optimizations. 

Previously, the test checks were assuming certain ordering of the terms, and that assumption was sometimes not met because of the compiler optimization.

Now, the updated test doesn't assume the ordering.

### DIFF
--- a/tests/test-disjointset.cpp
+++ b/tests/test-disjointset.cpp
@@ -57,22 +57,21 @@ static bool disjoint_set_rank(const Term & t1, const Term & t2)
 
 TEST_P(DisjointSetTests, TestDisjointSet)
 {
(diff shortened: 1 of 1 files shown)


yoni206 (approved):

yoni206:
This PR solves one of the two issues with CI. The other one should be fixed by https://github.com/stanford-centaur/smt-switch/pull/313 .
I am merging this PR, and then if the other one passes CI, we can merge it as well.

makaimann:
Thanks @ahmed-irfan!

yoni206:
Thanks @ahmed-irfan !

PR #318, opened 2022-11-22
Gtest upgrade

yoni206:
1. The main contribution of this PR is to upgrade the version of gtest that we are using. The motivation: I was unable to compile smt-switch on ubuntu, because of a gtest error.

2. The main consequence of the new gtest is that by default, empty test suites are not allowed and cause failures.
The solution for this is to allow empty test suites specifically for each test separately.
(We sometimes have empty test suites (e.g., tests for integers while linking only to bitwuzla).)

3. This way I also found some test suites that were unintentionally empty, and so tests didn't run. This is now fixed.

4. One of these tests, which seemingly didn't run, is now failing. The failed line is now commented out:
https://github.com/stanford-centaur/smt-switch/compare/master...yoni206:gtest_upgrade?expand=1#diff-421e4920f63052c195b027cfcfd2c01599e9663b5aca6e6e38dceed764d69332R87

If we merge this, I can create an issue to solve it (I don't know currently how).

6. I got another compiling error for not having `<limits>` included in `cvc5_solver`. This is fixed.

7. There was an empty test suite (no tests were defined in it) which i deleted: UnitQuantifierTests


src/smtlibparser.yy
```diff
@@ -268,7 +268,7 @@ term_s_expr:
          $$ = drv.solver()->make_term(po, *$3);
        }
     }
-    else if (uf = drv.lookup_symbol($2))
```
makaimann, line 271: Just a question for my understanding, is there a reason for the extra set of parentheses here or is that just a style thing?

makaimann (approved):
Thank you for working on this! Looks good to me. I only left one minor question, which is just for my understanding. If you could make an issue for that one line, that would be great. I can try to take a look at it soon.

src/smtlibparser.yy
yoni206: So this, and a couple of other unrelated diffs appear in this PR for some reason, even though they were already merged by Ahmed in his PR:
https://github.com/stanford-centaur/smt-switch/commit/53265962d835ecf88a9f59bd6e70c18f7056cdd8

The reason he added these was compiler warnings.

PR #323, opened 2023-01-19
Suppress printout from importing skbuild in configure script

cterrill26:
In response to issue [322](https://github.com/stanford-centaur/smt-switch/issues/322).

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -33,10 +33,11 @@ check_python_module("skbuild" "scikit-build")
 
 # Find cmake modules distributed with scikit-build
 # They are distributed under <scikit-build directory>/resources/cmake
+# Prevent printout when importing skbuild by temporarily redirecting stdout
 execute_process(
   COMMAND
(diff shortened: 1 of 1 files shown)


python/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ check_python_module("skbuild" "scikit-build")
 execute_process(
   COMMAND
     ${PYTHON_EXECUTABLE} -c "from __future__ import print_function; \
-      import os; import skbuild; \
+    import os; import sys; temp = sys.stdout; sys.stdout = open(os.devnull,\"w\"); import skbuild; sys.stdout = temp; \
```
yoni206, line 39: Thanks for this PR! Will you be able to add a comment above this line that explains why we are doing this? something like "importing os and skbuild, while suppressing warnings from skbuild"?

yoni206 (changes requested):

cterrill26:
Added the comment

yoni206 (approved):
Thanks for doing this!

PR #325, opened 2023-03-06
Added bitwuzla to pysmt frontend

jack-melchert:
Added bitwuzla to pysmt frontend, existing python tests are sufficient to test this.

### DIFF
--- a/python/smt_switch/pysmt_frontend/pysmt_solver.py
+++ b/python/smt_switch/pysmt_frontend/pysmt_solver.py
@@ -199,6 +199,22 @@ if 'btor' in  ss.solvers:
     SWITCH_SOLVERS['btor'] = SwitchBtor
 
 
+if 'bitwuzla' in  ss.solvers:
+    logics_params = dict(
+        quantifier_free=[True],
+        arrays=[True, False],
+        bit_vectors=[True],
+        uninterpreted=[True, False],
+        floating_point=[True, False],
+    )
+
+    class SwitchBitwuzla(_SwitchSolver):
+        LOGICS = _build_logics(logics_params)
+        _create_solver = ft.partial(ss.create_bitwuzla_solver, False)
+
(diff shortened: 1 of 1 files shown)


python/smt_switch/pysmt_frontend/pysmt_solver.py
```diff
@@ -199,6 +199,22 @@ def _reset_assertions(self):
     SWITCH_SOLVERS['btor'] = SwitchBtor
 
 
+if 'bitwuzla' in  ss.solvers:
+    logics_params = dict(
+        quantifier_free=[True],
+        arrays=[True, False],
```
yoni206, line 205: What does `[True, False]` mean in this context? Same for `uninterpreted` and `floating_point` below?

jack-melchert: Basically this is just an easy way of generating all of the supported logics for the solver. The list for each `logics_params `entry is the available settings for that option, with the default being [False]. Every possible combination of these options is then used to generate the list of supported logics.
For example, `quantifier_free=[True], bit_vectors=[True], uninterpreted=[True, False]` would result in supported logics of `QF_BV, QF_UFBV`.

yoni206: So what does the `False` mean in the `[True,False]` part of, say, `uninterpreted`?

jack-melchert: We want to generate every supported logic for this solver, which in this case includes both QF_BV and QE_UFBV, the `[True,False]` for uninterpreted says that the solver supports both. We are generating the list of supported logics from the every possible combination of these True/False lists.

yoni206 (approved):

jack-melchert:
@yoni206 would you be able to merge this? I don't have write access to smt-switch.

PR #336, opened 2023-11-15
Increase minimum CMake version

CyanoKobalamyne:
With the current version, CMake prints this warning: `Compatibility with CMake < 3.5 will be removed from a future version of CMake. Update the VERSION argument <min> value or use a ...<max> suffix to tell CMake that the project does not need compatibility with older versions.`


CyanoKobalamyne:
Since CMake 3.5 was released in December 2018 and every currently supported distro has newer versions, I'm merging this.

PR #337, opened 2023-11-15
Allow using local GTest

CyanoKobalamyne:
This searches for the system-provided version of GTest first, and only downloads it if it's not available.

Another possible improvement would be to not re-download GTest every time `configure.sh` is run.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -28,6 +28,7 @@ Configures the CMAKE build environment.
 --debug                 build debug with debug symbols (default: off)
 --static                create static libaries (default: off)
 --without-tests         build without the smt-switch test suite (default: off)
+--no-system-gtest       do not use system GTest sources; forces download (default: off)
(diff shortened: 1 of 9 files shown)


yoni206 (commented):
Looks good and will speed things up.
Two questions:
1. Perhaps it should verify that the local GTest is of the correct version, and only then avoid from downloading it?
2. What do you think about having an option for this PR? The default could be the new behavior, but it would be nice when running `configure.sh` to force a new download of gtest.

Both questions arise from painful previous experience with Gtest versions. And if that happens, or anything else goes wrong, it would be nice to just change an option in order to make sure that the right version is downloaded.

CyanoKobalamyne:
@yoni206 as per your requested, added version check and flag to force downloading. :)

yoni206 (approved):
Thanks for this!

PR #339, opened 2023-12-04
uint is non-standard

ffrohn:
This prevents compilation of smt-switch with alternative libc implementations like musl.


yoni206 (approved):
LGTM, thanks

PR #341, opened 2023-12-11
multiplication is variadic

ffrohn:
This allows for creating multiplication-terms with more than two arguments with Z3. Currently, such terms can be created with CVC5, but not with Z3 (I didn't try other solvers).

### DIFF
--- a/tests/test-int.cpp
+++ b/tests/test-int.cpp
@@ -41,6 +41,27 @@ class IntTests : public ::testing::Test,
   Sort intsort;
 };
 
+// test for creating variadic mult-terms
+TEST_P(IntTests, Mult)
+{
+  Term two = s->make_term(2, intsort);
+  Term three = s->make_term(3, intsort);
+  Term four = s->make_term(4, intsort);
+  Term twentyfour = s->make_term(24, intsort);
+  Term mult;
+  try
+  {
+    mult = s->make_term(Mult, {two, three, four});
+  }
+  catch (const IncorrectUsageException &e)
+  {
+    FAIL() <<  "creating mult-term failed: " << e.what();
+  }
+  s->assert_formula(s->make_term(Equal, twentyfour, mult));
+  auto res = s->check_sat();
+  assert(res.is_sat());
+}
+
 TEST_P(IntTests, IntDiv)
 {
   Term five = s->make_term(5, intsort);
--- a/include/ops.h
+++ b/include/ops.h
@@ -196,7 +196,7 @@ namespace std
 namespace smt {
 // ops that can be applied to n arguments
 const std::unordered_set<PrimOp> variadic_ops(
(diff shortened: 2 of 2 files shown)


yoni206:
Thanks! Could you please include a test for this new feature?
Perhaps the best place would be to add a test here?
https://github.com/stanford-centaur/smt-switch/blob/master/tests/test-int.cpp

This would also clarify the status with the other solvers.

tests/test-int.cpp
```diff
@@ -41,6 +41,27 @@ class IntTests : public ::testing::Test,
...
+  Term two = s->make_term(2, intsort);
+  Term three = s->make_term(3, intsort);
+  Term four = s->make_term(4, intsort);
+  Term twentyfour = s->make_term(24, intsort);
+  Term mult;
+  try
+  {
+    mult = s->make_term(Mult, {two, three, four});
+  }
+  catch (const IncorrectUsageException &e)
```
yoni206, line 56: Perhaps we could try this without catching? This will tell us what solvers have problems with that, and then we can fix it for those solvers?

Ideally we would like to be uniform w.r.t. back-end solvers, as much as possible.

ffrohn: I'm not sure what you mean. If an exception is thrown, then the FAIL in the catch-block is executed, and the solver's error message is shown:

`FAIL() << "creating mult-term failed: " << e.what();`

For example, if I execute the test in the master branch on my machine, then I get:

```
$SMT_SWITCH/tests/test-int.cpp:58: Failure
Failed
creating mult-term failed: Can't apply * to 3 terms.
[  FAILED  ] ParameterizedSolverIntTests/IntTests.Mult/4, where GetParam() = Z3(logging=0) (1 ms)
```

Since all CI checks pass with my fix, I guess that Z3 is the only solver that can't handle variadic multiplication.

I guess it would also work as desired without catching (I'm not very familiar with gtest and how it handles exceptions), but then we cannot provide a custom error message ("creating mult-term failed: "...).

yoni206: Sorry, you are right, I missed the fact that the message goes to `FAIL()`. So if `z3` is the only solver for which this does not work, could you perhaps try to make it work for that too? It should support it so it is unclear why this does not work:

https://github.com/stanford-centaur/smt-switch/blob/f4bd379333f06955eb147ac419b8d1f3fdd0d6fd/z3/src/z3_solver.cpp#L99

ffrohn: With my fix (a757315), it does work with Z3 (which is the reason why the CI checks passed after my last commit 61b4855). The problem was that `z3_solver` only tries to build a variadic term if the corresponding operation is marked as variadic in `ops.h`, which was not the case for `Mult`.

yoni206 (approved):
Thanks for this change and for the discussion!

PR #343, opened 2024-02-06
Set errexit bash option to terminate contrib setup scripts on error

CyanoKobalamyne:
Otherwise they continue even if a compilation step fails.


barrettcw (approved):
LGTM

CyanoKobalamyne:
It looks like this now causes CI to fail. I filed #344 which hopefully fixes this.

PR #346, opened 2024-03-12
Add datatype support for z3

farmerzhang1:
Hi all,

I'm trying to play with smt-switch and decide to add z3 backend support for datatypes. Any feedback is welcome.

### DIFF
--- a/contrib/setup-z3.sh
+++ b/contrib/setup-z3.sh
@@ -1,12 +1,20 @@
 #!/bin/bash
 set -e
 
-Z3_VERSION=6cc52e04c3ea7e2534644a285d231bdaaafd8714
+Z3_VERSION=z3-4.12.6
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
 mkdir -p $DEPS
+
+if [ "$(uname)" == "Darwin" ]; then
+    NUM_CORES=$(sysctl -n hw.logicalcpu)
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+    NUM_CORES=$(nproc)
+else
+    NUM_CORES=1
+fi
 
 if [ ! -d "$DEPS/z3" ]; then
     cd $DEPS
@@ -18,10 +26,20 @@ if [ ! -d "$DEPS/z3" ]; then
     # a pthread related issue
     # compiling with --single-threaded helps, but isn't a real solution
     # see https://github.com/Z3Prover/z3/issues/4554
-    ./configure --staticlib --single-threaded
-    cd build
-    make -j$(nproc)
+    # ./configure --staticlib --single-threaded
+    cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DZ3_BUILD_LIBZ3_SHARED=Off -DZ3_BUILD_LIBZ3_MSVC_STATIC=On
+    cmake --build build -j$NUM_CORES
     cd $DIR
 else
     echo "$DEPS/z3 already exists. If you want to rebuild, please remove it manually."
+fi
+
+if [ -f $DEPS/z3/build/libz3.a ]; then
+    echo "It appears z3 was setup successfully into $DEPS/z3."
+    echo "You may now install it with make ./configure.sh --z3 && cd build && make"
+else
+    echo "Building z3 failed."
+    echo "You might be missing some dependencies."
+    echo "Please see their github page for installation instructions: https://github.com/z3/z3"
+    exit 1
 fi
--- a/contrib/setup-z3.sh
+++ b/contrib/setup-z3.sh
@@ -26,7 +26,6 @@ if [ ! -d "$DEPS/z3" ]; then
     # a pthread related issue
     # compiling with --single-threaded helps, but isn't a real solution
     # see https://github.com/Z3Prover/z3/issues/4554
-    # ./configure --staticlib --single-threaded
     cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DZ3_BUILD_LIBZ3_SHARED=Off -DZ3_BUILD_LIBZ3_MSVC_STATIC=On
     cmake --build build -j$NUM_CORES
     cd $DIR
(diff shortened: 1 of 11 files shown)


contrib/setup-z3.sh
```diff
@@ -1,13 +1,21 @@
...
 mkdir -p $DEPS
 
+if [ "$(uname)" == "Darwin" ]; then
+    NUM_CORES=$(sysctl -n hw.logicalcpu)
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+    NUM_CORES=$(nproc)
+else
+    NUM_CORES=1
+fi
+
```
yoni206, line 18: I am not sure about this, perhaps this can go to a separate PR, while giving the user the option to choose number of cores?

contrib/setup-z3.sh
```diff
@@ -18,10 +26,20 @@ if [ ! -d "$DEPS/z3" ]; then
     # a pthread related issue
     # compiling with --single-threaded helps, but isn't a real solution
     # see https://github.com/Z3Prover/z3/issues/4554
-    ./configure --staticlib --single-threaded
-    cd build
-    make -j$(nproc)
+    # ./configure --staticlib --single-threaded
+    cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DZ3_BUILD_LIBZ3_SHARED=Off -DZ3_BUILD_LIBZ3_MSVC_STATIC=On
+    cmake --build build -j$NUM_CORES
```
yoni206, line 31: Is this change necessary? If so, I would delete the commented out line.

contrib/setup-z3.sh
```diff
@@ -18,10 +26,20 @@ if [ ! -d "$DEPS/z3" ]; then
...
+
+if [ -f $DEPS/z3/build/libz3.a ]; then
+    echo "It appears z3 was setup successfully into $DEPS/z3."
+    echo "You may now install it with make ./configure.sh --z3 && cd build && make"
+else
+    echo "Building z3 failed."
+    echo "You might be missing some dependencies."
+    echo "Please see their github page for installation instructions: https://github.com/z3/z3"
+    exit 1
+fi
```
yoni206, line 45: same question -- is this necessary? What is the current behavior?

z3/include/z3_datatype.h
```diff
@@ -0,0 +1,77 @@
+#pragma once
+#include <memory>
+#include <string>
+#include <vector>
```
yoni206, line 4: Are these needed? I don't see them in cvc5_datatype.h.

z3/include/z3_datatype.h
```diff
@@ -0,0 +1,77 @@
...
+#include "exceptions.h"
+#include "z3++.h"
+
+namespace smt {
+
+// forward declaration
+class Z3Solver;
+class Z3Datatype;
+class Z3DatatypeConstructorDecl;
+class Z3DatatypeDecl : public AbsDatatypeDecl
```
yoni206, line 16: Same question -- I don't see forward declarations in the cvc5 case. Can we try to make the z3 and cvc5 cases uniform? Or is that really needed only for z3?

z3/include/z3_datatype.h
```diff
@@ -0,0 +1,77 @@
...
+{
+ public:
+  Z3DatatypeDecl(z3::context & c, std::string name) : c(c), name(name){};
+
+ protected:
+  friend class Z3Solver;
+  friend class Z3Datatype;
```
yoni206, line 23: same

z3/src/z3_datatype.cpp
```diff
@@ -0,0 +1,40 @@
+#include "z3_datatype.h"
```
yoni206, line 1: This file is also pretty different from the cvc5 version, which is way shorter. Is that needed?

yoni206 (commented):
Thank you very much for this PR. It would be great to have support for z3 datatypes!
I've made some initial comments/questions inline.

In addition to them -- I wonder -- would it make sense to add tests that are specific to z3 datatypes? Perhaps here?
https://github.com/stanford-centaur/smt-switch/tree/main/tests/z3

We are already in good shape because I can see that CI ran the general datatypes tests, but I still wonder whether there is anything specific to test in the z3 API side.

contrib/setup-z3.sh
farmerzhang1: Well, I am copying this from contrib/setup-cvc5.sh

contrib/setup-z3.sh
farmerzhang1: Sure, I'll remove the comment

contrib/setup-z3.sh
farmerzhang1: Copied from contrib/setup-cvc5.sh as well.

z3/include/z3_datatype.h
farmerzhang1: Removed

z3/include/z3_datatype.h
farmerzhang1: Removed

z3/src/z3_datatype.cpp
farmerzhang1: Done. I kept only the compare function and the auxiliary functions for it.

z3/include/z3_datatype.h
farmerzhang1: Removed

farmerzhang1:
Hi Yoni, I've add some z3-specific tests and try to mirror the cvc5 datatype files as much as possible. Maybe you can check them again?

contrib/setup-z3.sh
```diff
@@ -18,10 +26,19 @@ if [ ! -d "$DEPS/z3" ]; then
     # a pthread related issue
     # compiling with --single-threaded helps, but isn't a real solution
     # see https://github.com/Z3Prover/z3/issues/4554
-    ./configure --staticlib --single-threaded
-    cd build
-    make -j$(nproc)
+    cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DZ3_BUILD_LIBZ3_SHARED=Off -DZ3_BUILD_LIBZ3_MSVC_STATIC=On
+    cmake --build build -j$NUM_CORES
```
yoni206, line 30: Why is this needed?

yoni206 (commented):
Thanks for the changes. Looks like the new tests are passing, and also the old ones, also after adding datatypes to the list of supported theories of z3, which is great.

Just one last question inline.

yoni206 (approved):
Thanks again for this valuable contribution!

PR #352, opened 2024-07-18
Unify static lib repacking to always use script

CyanoKobalamyne:
Use the `repack-static-libs.sh` script for repacking all static libs, because it can handle the archive files containing both x86_64 and arm64 code on macOS properly.


CyanoKobalamyne:
Replacement for #351, where I could not figure out how to change the base branch to `main`.

PR #353, opened 2024-08-28
Update bitwuzla version to bitwuzla/bitwuzla@0e81e616af4d44217298841928b194c3536c76

CyanoKobalamyne:


yoni206:
Thanks for this, LGTM. One thing that I wonder -- do we want to use latest commits of solvers, or only releases?

CyanoKobalamyne:
But I'm actually waiting for a fix to be pushed, so maybe hold off on merging this.

CyanoKobalamyne:
Actually, it seems that the fix will take some time, so I think that we should merge this now if we can. There have still been a lot of changes since the last update.

yoni206 (approved):

PR #355, opened 2024-10-08
Support all currently existing bitwuzla options

CyanoKobalamyne:


yoni206:
Thanks for this.
I am not sure however how realistic it is to keep track on all options of all solvers when updating their versions.
Can we, instead, do here what we already do for, e.g., cvc5?

https://github.com/stanford-centaur/smt-switch/blob/5b313f496718d6a786192dee39632c65f61ce1d0/cvc5/src/cvc5_solver.cpp#L134

That is, just trying to set the option as given and complain if it does not work? Would any `bitwuzla` option fail this way?

CyanoKobalamyne:
> That is, just trying to set the option as given and complain if it does not work? Would any `bitwuzla` option fail this way?

Okay, I updated the code to do just that.

yoni206 (approved):
Looks great, thanks!

PR #356, opened 2024-10-08
Add export pragmas to smt.h

CyanoKobalamyne:
This stops clangd from complaining about unused includes.

### DIFF
--- a/include/smt.h
+++ b/include/smt.h
@@ -17,26 +17,25 @@
 #pragma once
 
 // Exceptions used by SMT API.
-#include "exceptions.h"
+#include "exceptions.h"  // IWYU pragma: export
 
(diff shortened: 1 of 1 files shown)


CyanoKobalamyne:
Also, if the intention is for users to not include the individual headers referenced in `smt.h`, then we should probably mark them as private.

https://clangd.llvm.org/design/include-cleaner#umbrella-headers

barrettcw:
Actually, I would be in favor of using these pragmas and also documenting each "internal" header file to say that this file should not be included directly by a user.  I think this seems like a better practice than disabling the warnings.  If you're worried someone will forget in the future, we could also add some text in a "developer's wiki" explaining this policy.

yoni206 (approved):

PR #362, opened 2024-10-30
Mark internal headers private

CyanoKobalamyne:
Follow up to #356.

Also removed some unused headers and fixed some formatting errors.

### DIFF
--- a/include/exceptions.h
+++ b/include/exceptions.h
@@ -14,6 +14,8 @@
 **
 **/
 
+// IWYU pragma: private, include "smt.h"
+
 #pragma once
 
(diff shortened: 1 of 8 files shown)


yoni206 (commented):
We don't really have a developer guide right? But maybe there is some meaningful place where this can be documented? Perhaps somewhere in the readme?

barrettcw:
Why not use the github wiki?  This could be a start of something that eventually becomes a developer's guide.

yoni206 (approved):
I guess documenting this in the wiki can be done separately.

PR #367, opened 2024-11-22
Added fix and test to term translation

rachelcleaveland:
Now handles escape sequences in string literals and extra quotations. Bug fix for https://github.com/stanford-centaur/smt-switch/issues/365


src/term_translator.cpp
```diff
@@ -450,7 +450,11 @@ Term TermTranslator::value_from_smt2(const std::string val,
     }
   }
   else if (sk == STRING){
-    return solver->make_term(val, false, sort);
+    std::string new_val = val;
+    if (val.length() > 1 && val.at(0) == '\"') {
+      new_val = new_val.substr(1,new_val.length()-2);
+    }
+    return solver->make_term(new_val, true, sort);
   }
```
barrettcw, line 458: I don't understand this fix.  It looks like it just strips the double-quote character from the front of the string and whatever character is at the end of the string.  Shouldn't you check whether the last character is '\"' too?  And what about escape characters in the middle?

CyanoKobalamyne (approved):

PR #373, opened 2025-03-06
Generate tag and draft release from VERSION file

makaimann:
Rather than generate a `VERSION` file based on a tag, we now use the `VERSION` file as the single source of truth version. When that is updated in `main`, a GitHub Action will push a new tag and create a draft release.

This is better because the `VERSION` file used to trail the release by one commit. Since the source code packaged in a release doesn't include git metadata, the build system would use the VERSION file which would be out of date.


CyanoKobalamyne:
Thinking a about this now, I'm actually a bit concerned that this creates a sort of circular dependency.

If we create a release based on current main, that will have the old version file, because the workflow is triggered once the release is created, and creates a commit that follows the release, no?

CyanoKobalamyne:
Maybe this should be the other way around, no? I.e., we can manually update the version in this file, and that triggers a new release and new wheels being built.

makaimann:
That’s true the version file does lag the release by one commit... Hmm we could just specify it in two places but then they need to be kept in sync. What do you mean the other way around? I don’t think Git can tag based on a version file can it?

makaimann:
Maybe we should just drop the VERSION file entirely and if there’s no git metadata available, it just uses 0.0.0 or something. What do you think?

makaimann:
Okay looked into it more and I think making a GitHub Action to trigger a tag/draft release shouldn't be too bad. I'll try to get a working version in this branch. Might be a couple days before I get to that just fyi. Good suggestion!

python/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ find_package(Python 3.9 REQUIRED COMPONENTS Interpreter Development.Module)
 include(CheckPythonModule)
 
 # Verify required Python packages are installed
+check_python_module("packaging")
```
makaimann, line 12: Drive-by fix to prevent a confusing failure if the `packaging` module is not installed. Then CMake would incorrectly conclude the version was not sufficient.

makaimann:
Nevermind, I had a chance to look at it tonight. I tested this a bit, but won't really know for sure until we try it and let it make the draft release.

Let me know what you think!

Imo this is _much_ better -- thanks for bringing that up. Especially because I realized the source code packaged for a release does _not_ contain git metadata so it would have used the `VERSION` file with the wrong version. I completely forgot how the GitHub release process worked when I wrote that previous workflow.

.github/workflows/auto-draft-release.yml
```diff
@@ -0,0 +1,57 @@
...
+        if: steps.check_tag.outputs.tag_exists == 'false'
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "actions@github.com"
+          git tag -a v${{ steps.version.outputs.version }} -m "Version ${{ steps.version.outputs.version }}"
+          git push origin "v${{ steps.version.outputs.version }}"
+
+      - name: Create draft GitHub Release
+        if: steps.check_tag.outputs.tag_exists == 'false'
+        uses: softprops/action-gh-release@v1
```
CyanoKobalamyne, line 49:
```suggestion
        uses: softprops/action-gh-release@v2
```
Maybe we should use v2; that seems to be the latest version: https://github.com/softprops/action-gh-release/releases.

.github/workflows/auto-draft-release.yml
```diff
@@ -0,0 +1,57 @@
...
+          else
+            echo "tag_exists=false" >> $GITHUB_OUTPUT
+          fi
+
+      # Create and push the tag if it doesn't already exist
+      - name: Create and push tag
+        if: steps.check_tag.outputs.tag_exists == 'false'
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "actions@github.com"
```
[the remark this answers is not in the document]
makaimann: Nice find! Sounds like that should improve the commit message icon.

makaimann:
> I'm actually surprised it works out of the box, because packaging is a [PyPI package](https://pypi.org/project/packaging/), not included with base Python (at least on my system).

Yeah my guess is I have it installed locally and it's included in the `manylinux` and Mac `cibuildwheel` environments by default.

python/CMakeLists.txt
makaimann: Yeah my guess is I have it installed locally and it's included in the `manylinux` and Mac `cibuildwheel` environments by default.

makaimann:
> why do you create a release draft, instead of making the release directly?
Ah I jus want a chance to modify the release notes.

CyanoKobalamyne (approved):

PR #384, opened 2025-04-17
Fix wheels CI for building wheels

makaimann:
There were two minor issues causing the `Build Wheels` CI job to fail. This PR resolves both.

### Issue 1
After https://github.com/stanford-centaur/smt-switch/pull/369 merged, the `setup-cvc5.sh` script could not find the `cvc5` libraries on aarch64, because it's stored in `<prefix>/lib64` instead of `<prefix>/lib`.

**Solution:** This PR simply checks both `lib` and `lib64`. This seemed simpler than trying to determine which path it should be based on the architecture -- either location is fine regardless.

### Issue 2
The `CIBW_TEST_COMMAND` is how `cibuildwheel` tests the final wheel. This test was failing nondeterministically. It was because `cibuildwheel` creates a fresh virtualenv that it installs the final (repaired and packaged) wheel in. However, using `pytest` on the CLI used the system python instead of the virtualenv python. Sometimes this found the wrong wheel (the one in the build dir before it's repaired).

**Solution:** I'm not sure why this sometimes worked and sometimes didn't, but ensuring we use the virtualenv python appears to resolve this issue.

### DIFF
--- a/contrib/setup-cvc5.sh
+++ b/contrib/setup-cvc5.sh
@@ -23,7 +23,8 @@ if [ ! -d "$DEPS/cvc5" ]; then
     git clone https://github.com/cvc5/cvc5.git
     cd cvc5
     git checkout -f ${CVC5_VERSION}
-    ./configure.sh --prefix=$DEPS/install --static --auto-download --dep-path="$DEPS/install"
+    # ensure the cvc5 libraries are placed in deps/install/lib
+    ./configure.sh --prefix=$DEPS/install --static --auto-download --dep-path="$DEPS/install" -DCMAKE_INSTALL_LIBDIR=lib
     cd build
     make -j$NUM_CORES
     make install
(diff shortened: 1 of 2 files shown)


CyanoKobalamyne (approved):
Thanks!

contrib/setup-cvc5.sh
CyanoKobalamyne, line 1: Actually, sorry, would you mind checking if adding `-DCMAKE_INSTALL_LIBDIR=lib` to the `./configure.sh` line also fixes this issue?

It would be great if we could rely on the libraries always being in `deps/install/lib`. I actually thought we already implicitly assume this in some places.

makaimann: Yep good call! That does the trick. Updated the PR.

CyanoKobalamyne (approved):

PR #389, opened 2025-04-30
Fix linking of z3 library

### LINKED ISSUE #388: `libsmt-switch-z3.a` is missing symbols
Linking the static smt-switch z3 library into a shared object results in "symbol not found" errors. This can be seen, e.g., when trying to use the Python C extension built as such: https://github.com/stanford-centaur/smt-switch/actions/runs/14526964832/job/41378477258?pr=382.

The error raised here is:
```
ImportError: dlopen(/Users/runner/work/smt-switch/smt-switch/build/python/smt_switch/smt_switch_base.cpython-313-darwin.so, 0x0002): symbol not found in flat namespace '__ZN11bit_blasterC1ER11ast_managerRK18bit_blaster_params'
```

The missing symbol is present in `libz3.a` but not in `libsmt-switch-z3.a` despite the fact that we "repack" the former into the latter. Upon closer inspection, the file this symbol is from (`src/ast/rewriter/bit_blaster/bit_blaster.cpp`) is not in the final archive. This is because there is another file with the same name (`src/ast/simplifiers/bit_blaster.cpp`) that overwrites it during extraction, since `ar` does not preserve paths, only filenames.

CyanoKobalamyne:
Fixes #388.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,9 +85,16 @@ jobs:
 
       - name: Build
         run: |
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+            NUM_CORES=$(nproc)
+          elif [[ "$RUNNER_OS" == "macOS" ]]; then
+            NUM_CORES=$(sysctl -n hw.ncpu)
+          else
+            NUM_CORES=2
+          fi
           source ./.venv/bin/activate
           cd build
-          make -j
+          make -j$NUM_CORES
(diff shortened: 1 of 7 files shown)


CyanoKobalamyne:
I cannot reproduce this error in a fresh Ubuntu 24.04 container.

CyanoKobalamyne:
Merging this, since it fixes at least some of the errors.

PR #397, opened 2025-05-10
Update cvc5 setup script

CyanoKobalamyne:
Makes it more similar to the other scripts and ensures that build tools other than `make` work.


contrib/setup-cvc5.sh
```diff
@@ -16,27 +17,22 @@ else
...
+    "$CONTRIB_DIR/setup-cadical.sh"
+    git clone https://github.com/cvc5/cvc5.git && cd cvc5
+    git checkout -f $CVC5_VERSION
     # ensure the cvc5 libraries are placed in deps/install/lib
     ./configure.sh --prefix=$DEPS/install --static --auto-download --dep-path="$DEPS/install" -DCMAKE_INSTALL_LIBDIR=lib
     cd build
-    make -j$NUM_CORES
-    make install
-    cd $DIR
+    cmake --build . --target install -j$NUM_CORES
```
CyanoKobalamyne, line 29: This is the main change.

yoni206: just wondering -- why shouldn't we install it according to the instructions of cvc5?

CyanoKobalamyne: I changed this so that we can use `ninja` instead of `make` transparently. I opened a PR in cvc5 too: https://github.com/cvc5/cvc5/pull/11879

yoni206: in that case, should we include `ninja` as a dependency?

CyanoKobalamyne: I added it to the `README`. Either Make or Ninja should work.

yoni206 (approved):

PR #409, opened 2025-06-04
Unify setup scripts

CyanoKobalamyne:
This PR creates a common format for the solver setup scripts in `contrib/`, inspired by Arch Linux's PKGBUILD. Now they essentially only need to:
1. define a couple of variables describing where they should be downloaded from,
2. define the `configure_step`, `build_step`, and `install_step` functions and/or source the `xxx-steps.sh` script corresponding to their build system (GNU autotools/make, CMake, Meson), or
3. source the `common-setup.sh` script that performs the setup common to all dependencies then executes the steps defined above.

This makes the scripts shorter, less error-prone, and easier to update.


CyanoKobalamyne:
@yoni206, do you have any thoughts about this?

yoni206 (approved):

PR #411, opened 2025-06-04
Compute interpolants incrementally in MathSAT

Po-Chun-Chien:
This PR improves the computation of interpolation sequences when using MathSAT.
The interpolator now attempts to reuse asserted formulas from the previous call on the solver stack, enabling incremental SMT solving.

An evaluation on 395 Btor2 BV tasks was done with https://github.com/stanford-centaur/pono/commit/a55bb6b0bc1e190820905af42b1ab3ebb3d6b3e4,
which contains the new algorithms and options introduced by https://github.com/stanford-centaur/pono/pull/381, https://github.com/stanford-centaur/pono/pull/400, and https://github.com/stanford-centaur/pono/pull/414.
The results are available in this [table](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-295-ga55bb6b/itp-engines.msat-incr.table.html#/).

Notably, with incremental SMT solving, the following engines show improvements in terms of the number of solved tasks:

-  `interp` engine with `--interp-backward` (https://github.com/stanford-centaur/pono/pull/381) improved by [4 %](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-295-ga55bb6b/itp-engines.msat-incr.table.html#/?hidden4=2,3,4&hidden9=2,3,4&hidden=0,1,3,4,5,6,8,9&hidden2=2,3,4&hidden7=2,3,4)
- `ismc` engine (https://github.com/stanford-centaur/pono/pull/400) improved by [12 %](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-295-ga55bb6b/itp-engines.msat-incr.table.html#/?hidden4=2,3,4&hidden9=2,3,4&hidden=0,1,2,3,5,6,7,8)
- `ic3ia` engine with `--no-ic3ia-sim-cex` (https://github.com/stanford-centaur/pono/pull/414) improved by [5 %](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-295-ga55bb6b/itp-engines.msat-incr.table.html#/?hidden4=2,3,4&hidden9=2,3,4&hidden=0,2,3,4,5,7,8,9&hidden2=2,3,4&hidden7=2,3,4&hidden1=2,4,3&hidden6=2,4,3)


Po-Chun-Chien:
An interesting and unexpected side effect of this PR:
Some segmentation faults previously caused by MathSAT no longer occur when running Pono's interpolation engine.
I suspect this might be related to the use of push/pop operations, which weren’t part of the previous implementation.

msat/src/msat_solver.cpp
```diff
@@ -1264,74 +1274,93 @@ Term MsatInterpolatingSolver::get_value(const Term & t) const
...
-Result MsatInterpolatingSolver::get_sequence_interpolants(
-    const TermVec & formulae, TermVec & out_I) const
-{
-  initialize_env();
-  // reset the environment -- each get_sequence_interpolants is it's own
-  // separate call
-  msat_reset_env(env);
+  // update the interpolation groups and assertions
+  assert(num_reused == msat_num_backtrack_points(env));
+  interp_grps_.resize(num_reused);
```
CyanoKobalamyne, line 1341: I'm not entirely sure what `msat_num_backtrack_points` returns, but won't this assertion fail if we only reuse a prefix of all existing assertions? Say, if the previous query was for the sequence interpolant of `A`, `B`, `C`, and this one is just `A`, `B`. In that case, `num_reused = 2`, but what will `msat_num_backtrack_points` return?

CyanoKobalamyne:
It would also be good to add or expand some tests for this functionality, to check all possible cases of query sequences.

msat/src/msat_solver.cpp
Po-Chun-Chien: My previous fix was incomplete. 
2632dcb should ensure that the assertion holds.

In the above example,
`msat_num_backtrack_points` will return 3 when entering `get_sequence_interpolants`,
followed by 1 `msat_pop_backtrack_point`,
then `msat_num_backtrack_points` will return 2 at this line.

CyanoKobalamyne (approved):

PR #413, opened 2025-06-16
Cvc5 api update

CyanoKobalamyne:
This PR updates uses of `Solver` methods to `TermManager` methods where the former have been marked as deprecated.

It also bring the BV to integer operations up to spec now that SMT-LIB 2.7 has been released and incorporated into cvc5.


CyanoKobalamyne:
@yoni206 It should all be fixed now.

cvc5/src/cvc5_term.cpp
```diff
@@ -225,7 +229,7 @@ Op Cvc5Term::get_op() const
   if (kind2primop.find(cvc5_kind) == kind2primop.end())
   {
     throw NotImplementedException("get_op not implemented for cvc5 Kind "
-                                  + cvc5::kindToString(cvc5_kind));
+                                  + std::to_string(cvc5_kind));
```
barrettcw, line 232: Does this still print the name of the kind?  Or just an integer?

yoni206: This is a good point -- maybe a test can be added to verify that the name of the kind is printed?

CyanoKobalamyne: I'm not sure. I can check, but `cvc5::kindToString` has been marked deprecated and will be removed in the next minor cvc5 release according to the cvc5 release notes, so I don't think we should continue using it.

CyanoKobalamyne: Based on the implementation here, it will print the name of the kind: `https://github.com/cvc5/cvc5/blob/58cda4cdc96a9cdc0e267ba4f4ddd00bcd653a1f/src/api/cpp/cvc5.cpp#L9018`. I think this is identical to the former `kindToString` function, they just changed the name to be more standard-conformant.

CyanoKobalamyne: Just verified this: it still prints the name of the sort kind.

PR #419, opened 2025-08-08
Fix symbol names in printing solver

### LINKED ISSUE #418: Printing solver prints wrong symbol name for some solvers
Certain solvers, like Boolector, add extra pre- or suffixes to symbol names when created via their programmatic API. In these cases, `PrintingSolver` puts the wrong name in the `define-fun` line.

CyanoKobalamyne:
Fixes #418.


src/printing_solver.cpp
```diff
@@ -151,24 +182,31 @@ Term PrintingSolver::make_term(const Term & val, const Sort & sort) const
...
+  {
     range_str = sort->to_string();
   }
-  (*out_stream) << "(" << DECLARE_FUN_STR << " " << name << " " << "(" << domain_str << ")" << " " << range_str << ")" << endl;
-  return wrapped_solver->make_symbol(name, sort);
+  Term sym = wrapped_solver->make_symbol(name, sort);
+  std::string name_str = sym->to_string();
+  (*out_stream) << "(" << DECLARE_FUN_STR << " " << name_str << " " << "("
+                << domain_str << ")" << " " << range_str << ")" << std::endl;
+  return sym;
```
CyanoKobalamyne, line 206: This is the actual change, everything else is formatting.

yoni206 (commented):
Thanks a lot for fixing this!
Any chance you could add the test from the issue?

PR #391, opened 2025-05-01
Support reset-assertions in Bitwuzla API

Po-Chun-Chien:
According to the [documentation](https://bitwuzla.github.io/docs/c/api.html#reset-assertions-example), one can reset assertions in Bitwuzla by discarding the current solver instance and creating a new one.

This feature allows us to use Bitwuzla as the base solver in the `interp` engine of Pono (see https://github.com/stanford-centaur/pono/pull/395#discussion_r2060967479).

PR #420, opened 2025-08-14
Fix some issues in  `reset_assertions` of `BzlaSolver`

Po-Chun-Chien:
See also #391.


Po-Chun-Chien:
The pipeline is green on my branch.

PR #428, opened 2025-09-25
Add support for Floating-Point theory (currently cvc5 only)

augustomafra:
Adding sort and operation definitions for Floating-Point support on smt-switch.

This is a first step for other upcoming changes on pono's SMV frontend encoding problems on the theory of Floating-Point arithmetic.


augustomafra:
Hi, @CyanoKobalamyne ,

Could you or other project maintainers share feedback about this PR?
I have merged it to the latest commit in main (version 1.0.6), so please let me know if there is anything else I can adjust on my request.
I'm currently working on my thesis to wrap up my Master's project at https://github.com/augustomafra/pono, so any thoughts from the Pono maintainers would be greatly appreciated, if possible.

Thanks!

include/ops.h
```diff
@@ -131,6 +131,39 @@ enum PrimOp
...
+  FPIsNeg,
+  FPIsPos,
+  IEEEBV_To_FP,
+  FP_To_FP,
+  Real_To_FP,
+  SBV_To_FP,
+  UBV_To_FP,
+  FP_To_UBV,
+  FP_To_SBV,
+  FP_To_REAL,
```
yoni206: Can you also add python bindings as mentioned in the line below?

yoni206 (commented):
Hi @augustomafra and thanks for this PR, and sorry for the delay!
It would be great for smt-switch to support floating points. 

In addition to the inline comments, could you add tests to the PR? I think it would make sense to add cvc5 tests here:
https://github.com/augustomafra/smt-switch/tree/63258f708b798a34a46a6ddaf7252ece4612af5b/tests/cvc5

and more generic tests (though they would now only run with cvc5) here:
https://github.com/augustomafra/smt-switch/tree/63258f708b798a34a46a6ddaf7252ece4612af5b/tests

This would include updating this file:
https://github.com/augustomafra/smt-switch/blob/63258f708b798a34a46a6ddaf7252ece4612af5b/tests/available_solvers.cpp

augustomafra:
Hi, @yoni206 ,

Sorry for the long time without updates. I'm working on the pull request now and added the general and cvc5 specific tests. The general tests cover the basic functioning of the FP operators, even though they only run with cvc5 now. I'm using the cvc5 specific tests to cover the special FP values such as rounding modes and NaN/Inf. 

Please let me know once you have some time to continue the review, and feel free to suggest other improvements :)

I'm adding other comments regarding the inline reviews as well.

include/sort.h
```diff
@@ -51,13 +51,53 @@ enum SortKind
...
+  static constexpr uint32_t size = exp + sig;
+};
+
+template <>
+struct FPSizes<FLOAT64>
+{
+  static constexpr uint32_t exp = 11;
+  static constexpr uint32_t sig = 53;
+  static constexpr uint32_t size = exp + sig;
+};
```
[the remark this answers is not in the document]
augustomafra: I don' think these templates are strictly needed, but it turned out that they were useful for avoiding hard coded numbers in my implementation: https://github.com/stanford-centaur/smt-switch/pull/428/changes#diff-9c5f10fab20b8842000e6991b238ab089c359600f93c6f4b75984a24d078f641R343-R352. At least for cvc5, I'm supporting only Float32 and Float64 values, so we need to check these sizes during term construction.

I preferred to keep the templated definitions as a way to make the API more extensible: we could implement FPSizes<16> and FPSizes<128> for other float types in the future.

I'm open to other implementations if you think it is fitting though :)

include/ops.h
augustomafra: I started working on this, but it became a more involved solution as it would be good to have python bindings to other FP enums such as ROUND_NEAREST_TIES_TO_EVEN, POS_INFINITY and NOT_A_NUMBER.

I will develop this change on a separate pull request (I promise! I have worked on other changes extending FP Smt-Switch support to MathSAT and Bitwuzla, so adding one more PR to the stack will be ok).

